PR #114, opened 2025-09-30
SPT-1620: Validate Identity and process `vtr`

chris-dco:
### What changed

* Upon successful verification of signature and determination that the Fraud credential has not expired, validate that the stored identity and correctly report the isValid and vot fields.

  To validate this will check the `vot` present in the identity satisfies the requested `vtr`.

* Updated OpenAPI specification to ensure that AWS API Gateway will validate the input by adding the missing `x-amazon-apigateway-request-validators.ALL`.

### Why did it change

Before returning an identity we first need to decide whether it is usable for the relaying parties request. We should not return an identity that has either been marked invalid, if the account has been deleted or, for which, the constituent VCs have changed.  Nor should we return a usable identity if the identity record does not meet the requested vtr.

### Issue tracking

- SPT-1620

## Testing

Added unit and integration tests.

### Manual Steps to Test

None

### Checklist for developers

- [x]  If there will be further PRs related to the story, this is highlighted in the PR description what is to come
- [x]  All commits are signed
- [x]  All commits contain a `Co-authored-by` line where pairing has taken place

### Checklist for reviewers


src/commons/calculate-si-is-valid.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type IsSiValidOutput = {
+  isValid: true;
+  vot: IdentityVectorOfTrust;
+} & {
+  isValid: false;
+  vot: "P0";
+};
+
+export const isSiValid = (vot: IdentityVectorOfTrust, vtr: string[], isDeleted: boolean = false): IsSiValidOutput => {
```
michaelabenyohai, line 11: After the discussion about the acceptance criteria for this story earlier, we decided not include the check for `isDeleted` and we also decided that this should not be setting the `isValid` value (the only thing that will set `isValid` now is SPT-1629 (checking the VC signatures match those in the stored identity).

Can you change this logic and rename the file/functions to be framed around VOT calculation only, and not on the concept of `isValid`?

openAPI/api.yaml
```diff
@@ -72,14 +72,8 @@ components:
...
-          type: array
-          items:
-            type: string
-            enum:
-              - P1
-              - P2
-              - P3
-              - P4
+          type: string
+          pattern: "^P[1-4](,P[1-4])*$"
```
michaelabenyohai: I assume this pattern works for if there are more than 2 PXs? That's what the `*` is for right?

src/commons/calculate-si-is-valid.ts
chris-dco: I have updated this and it now only returns the `vot`. I have also moved the function.

openAPI/api.yaml
michaelabenyohai: Ah ok, maybe just squash it all into a single commit? There's not a huge amount of change overall and a lot more change across the 5 commits that overall gets changed/reverted. Would squashing the whole lot work?

src/identity-reuse/tests/calculate-vot.test.ts
```diff
@@ -0,0 +1,23 @@
...
+  it.each<[string, IdentityVectorOfTrust[], string]>([
+    ["P1", ["P1"], "P2"],
+    ["P2", ["P1", "P2"], "P2"],
+    ["P1", ["P1", "P2"], "P1"],
+    ["P2", ["P2"], "P3"],
+    ["P2", ["P2", "P3"], "P2"],
+    ["P2", ["P2"], "P3"],
+    ["P3", ["P2", "P3"], "P3"],
+    ["P2", ["P2", "P3"], "P2"],
+    ["P0", ["P3"], "P2"],
```
michaelabenyohai, line 14: Can you add some examples where we have multiple items in the VTR, but reverse the order? The implementation currently has a line of code that makes sure that we always get the same result no matter the order of the VTR, but this test doesn't test that. So, for example, have both:

- `["P2", ["P1", "P2"], "P2"]` (exists)
- `["P2", ["P2", "P1"], "P2"],` (does not exist)

chris-dco: Added

PR #265, opened 2026-04-10
SPT-1902: Create SIS endpoints

chris-dco:
### What changed

* Created a new PUBLIC API Gateway (for non-production environment only):
  * Adds the new stub `/authorize` endpoint which returns a hard coded `code` which can be sent to the `/token` endpoint in exchange for an access token.
* Created a new PRIVATE API Gateway (for non-production environment only):
  * Adds a new stub `/token` endpoint which will exchange a `code` provided by the `/authorize` endpoint with an access token.
  * Added new stub `/user-identity` endpoint, which will eventually replace the existing user-identity endpoint.
* Created a new REGIONAL API Gateway which is a duplicaate of the PRIVATE Gateway for testing only in the dev environment.
* Created API specification documents for each of the new API Gateways.
* Added unit and acceptance tests for each of the APIs.

### Why did it change

SIS Phase 3 will enable Orchestration to consume SIS directly via OAuth and determine whether a user can reuse an existing identity without invoking IPV Core.

This ticket is required as part of the initial skeleton work for Phase 3, specifically building the endpoints which will exist within SIS and enable the OAuth flow.

## Testing

* New unit tests
* New acceptance tests

### Manual Steps to Test

N/A

### Checklist for developers

- [ ]  There are no merge, WIP or reversion commits included in the PR (reversion commits allowed if reverting previously merged code)
- [ ]  All commits include story reference, a distinct title and a body listing changes
- [ ]  All commits are signed
- [ ]  All changes in this PR are related to the story
- [ ]  The changes have been fully tested in the dev environment
- [ ]  There are unit and/or integration tests matching story acceptance criteria (where applicable)

### Checklist for reviewers


template.yaml
```diff
@@ -237,87 +237,938 @@ Conditions:
...
+    Type: AWS::Serverless::Api
+    Condition: IsNotProduction
+    Properties:
+      # checkov:skip=CKV_AWS_120: API Gateway caching not enabled
+      Name: !Sub "${AWS::StackName}-PublicApi"
+      AlwaysDeploy: true
+      OpenApiVersion: 3.0.3
+      PropagateTags: true
+      DefinitionBody:
+        openapi: "3.0.3" # workaround to get `sam validate` to work
```
joe-mcgarry, line 912: Is this (and the same in `privateRestApi`) only flagged as a workaround for the short term until things are more fleshed out? Or does it not matter as it's the regional API just for testing (I think)?

joe-mcgarry (commented):
One minor comment but looks good otherwise

openAPI/private-api.yaml
```diff
@@ -0,0 +1,94 @@
+---
+openapi: 3.0.3
+info:
+  title: "SIS Private – Stored Identity"
```
michaelabenyohai: Small nice-to-have. The title of the private APi is just `SIS Public API` and I wonder if we should just call this one `SIS Private API`. The additional "Stored Identity" doesn't add much and it is nice to be a bit consistent across the two titles so that we can tell what the key difference is.

template.yaml
chris-dco: It's a general bug with SAM validation. If you search around other repos you'll find it all over.

openAPI/private-api.yaml
chris-dco: Thanks. I have updated it.

openAPI/public-api.yaml
```diff
@@ -0,0 +1,118 @@
...
+      required: false
+      description: |
+        A space-separated list of scopes which must include the `openid` value. Other options are `email`, `phone`
+        and `offline_access` (returns a refresh token).  Other custom scopes are available for internal requests.
+      schema:
+        type: string
+    StateParam:
+      in: query
+      name: state
+      required: false
```
michaelabenyohai: I wonder if we should set this to `true`. The OAuth RFC says it is RECOMMENDED so not technically required, but RFC-0001 detailing the GOV.UK One Login OAuth profile mentions that clients MUST use the `state`.

chrisclayson: Agreed ... I probably should have linked to RFC-0001 in the story 🤦🏻

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+          Fn::GetAtt: UserIdentityHandlerRestApiRole.Arn
+        passthroughBehavior: NEVER
+        contentHandling: CONVERT_TO_TEXT
+        type: AWS_PROXY
+
+components:
+  schemas:
+    TokenRequestBody:
+      type: object
+      properties:
```
michaelabenyohai, line 100: ℹ️  Just a note to as a reminder for us, though I think these are things for future stories (not for this PR).

We will need to add to these properties:

* `redirect_uri` - we need to write a story to check this is the same value as provided in the authorization request as per [rfc6749](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3)
* `client_id` and other properties for client authentication using [private_key_jwt](https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication) - let's write a story for implementing client authentication

src/handlers/authorize-handler/tests/authorize-handler.test.ts
michaelabenyohai, line 1: File name has a typo, should be `authorize-handler.test.ts`

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -0,0 +1,48 @@
...
+    redirect_uri: redirectUri,
+    response_type: responseType,
+    scope,
+  } = event.queryStringParameters as AuthorizationQueryStringParameters;
+
+  if (responseType !== "code") {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        message: "Request parameter [response_type] must be of type 'code'",
```
michaelabenyohai: Is it worth adding a `log.error` here as well? I wonder if that would be useful for our own debugging.

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  const {
+    redirect_uri: redirectUri,
+    response_type: responseType,
+    scope,
+  } = event.queryStringParameters as AuthorizationQueryStringParameters;
+
+  if (responseType !== "code") {
+    return {
+      statusCode: 400,
```
michaelabenyohai: I believe this scenario should not return a `400`. According to [RFC 6749 section 4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) this should be an OAuth "error response", which means it returns a `302` to the usual redirect uri with an error code `unsupported_response_type`.

```
HTTP/1.1 302 Found
Location: https://client.example.com/cb?error=unsupported_response_type&state=xyz
```

I wonder if this sort of error validation is out of scope for this ticket and we should just remove the validation and associated `400` from the openapi spec. We can implement it in later stories as there are lots of other error scenarios we need to cover.

src/handlers/authorize-handler/authorize-handler.ts
michaelabenyohai: Might not actually be necessary based on other comment that suggests removing this code.

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -0,0 +1,48 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult, Context } from "aws-lambda";
+import logger from "../../commons/logger";
+import { URL } from "node:url";
+
+export type OAuthBadRequest = {
+  error: string;
+  error_description: string;
+};
+
+export type AuthorizationQueryStringParameters = {
```
michaelabenyohai, line 10: I think `state` is missing here?

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -0,0 +1,48 @@
...
+      body: JSON.stringify({
+        message: "Request parameter [response_type] must be of type 'code'",
+      }),
+    };
+  }
+
+  const url = new URL(decodeURIComponent(redirectUri));
+  url.searchParams.append("code", "SplxlOBeZQQYbYS6WxSbIA");
+  if (scope) {
+    url.searchParams.append("scope", scope);
```
michaelabenyohai: I think this should be `state` rather than `scope`?

src/handlers/token-handler/token-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+  logger.addContext(context);
+
+  if (!event.headers["content-type"] || event.headers["content-type"] !== "application/x-www-form-urlencoded") {
+    logger.error(
+      `'content-type' is '${event.headers["content-type"]}' but should be 'application/x-www-form-urlencoded'`
+    );
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        error: "invalid_type",
```
michaelabenyohai: According to [RFC 6749 (section 5.2)](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2)  I think this should be `invalid_request`? `invalid_type` is not one of the listed error codes.

src/handlers/token-handler/token-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+      }),
+    };
+  }
+
+  const searchParams = new URLSearchParams(event.body || "");
+  if (searchParams.get("grant_type") !== "authorization_code") {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        error: "invalid_field",
```
michaelabenyohai: According to RFC 6749 (section 5.2) I think this should be `invalid_grant`?

src/handlers/token-handler/token-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+      }),
+    };
+  }
+
+  const code = searchParams.get("code");
+  if (!code || code.length <= 0) {
+    return {
+      statusCode: 400,
+      body: JSON.stringify({
+        error: "invalid_field",
```
michaelabenyohai: According to RFC 6749 (section 5.2) I think this should be `invalid_grant`?

src/handlers/token-handler/token-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+        error_description: `missing field [code]`,
+      }),
+    };
+  }
+
+  return {
+    statusCode: 200,
+    body: JSON.stringify({
+      access_token: createHash("sha256").update(code).digest("hex"),
+      scope: "unknown",
```
michaelabenyohai: Given that `scope` is optional (and we're probably unlikely to support it) I wonder if we should just not return it?

src/handlers/token-handler/token-handler.ts
```diff
@@ -0,0 +1,53 @@
...
+      token_type: "example",
```
michaelabenyohai: I think in our case this is `"Bearer"` - RFC 0001 mentions that we issue "bearer" access tokens which is described further in [RFC 6750 (OAuth 2.0: Bearer Token Usage)](https://datatracker.ietf.org/doc/html/rfc6750#section-4)

src/handlers/authorize-handler/authorize-handler.ts
chris-dco: Changing this to state.

src/handlers/authorize-handler/authorize-handler.ts
chris-dco: Code removed.

src/handlers/authorize-handler/authorize-handler.ts
chris-dco: Code removed.

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+      summary: "Retrieve the user’s stored identity record"
+      description: >
+        Returns the signed Stored Identity JWT and associated metadata
+        indicating whether the record is still valid, whether it has expired,
+        and the stored vector of trust (VOT).
+      security:
+        - bearerAuth: []
+      tags:
+        - Stored Identity
+      requestBody:
```
michaelabenyohai: My understanding of the new phase-3 version of the `/user-identity` endpoint is that it will not have a `requestBody` at all. Instead this information will be provided to the new `/authorize` endpoint and the new `/user-identity` endpoint simply uses the bearer token to get the required information from the session.

It will be the same as the IPV Core version of [this endpoint](https://github.com/govuk-one-login/ipv-core-back/blob/894dce504b846958c65575cb665f4da0fc15be9d/openAPI/core-back-external.yaml#L48).

Can you remove the `requestBody` from here?

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+      x-amazon-apigateway-integration:
+        httpMethod: POST
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${TokenFunction.Arn}:live/invocations
+        credentials:
+          Fn::GetAtt: TokenFunctionRestApiRole.Arn
+        passthroughBehavior: never
+        type: AWS_PROXY
+  /user-identity:
+    post:
```
michaelabenyohai: My understanding of the new phase-3 version of the `/user-identity` endpoint is that this should be a `get` since (as per other comment) there will be no request body. This is the same as the current [IPV Core version of this endpoint](https://github.com/govuk-one-login/ipv-core-back/blob/894dce504b846958c65575cb665f4da0fc15be9d/openAPI/core-back-external.yaml#L48).

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+      summary: "Retrieve the user’s stored identity record"
+      description: >
+        Returns the signed Stored Identity JWT and associated metadata
+        indicating whether the record is still valid, whether it has expired,
+        and the stored vector of trust (VOT).
```
michaelabenyohai: I think this should just be "Returns the signed stored identity and vector of trust (VOT)".

The new version of the endpoint won't return all the "expired" metadata, as that will be reflected in OAuth error responses from the new `/authorize` endpoint in the phase 3 world.

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UserIdentityInput"
+      responses:
+        "200":
+          description: "Stored identity record returned"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/StoredIdentityResponse"
```
michaelabenyohai: My understanding of the new phase-3 version of the `/user-identity` endpoint is that it will be returning `#/components/schemas/UserIdentityDataType` and not the `#/components/schemas/StoredIdentityResponse`.

It will only return the result, and none of the "isExpired" metadata - the expiry stuff will instead be reflected in the OAuth error responses in the new `/authorize` endpoint in future.

openAPI/private-api.yaml
```diff
@@ -0,0 +1,278 @@
...
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OAuthError"
+        "403":
+          description: "Forbidden – e.g. expired access token"
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/OAuthError"
+        "404":
```
michaelabenyohai: I don't think we need the `404` response. The use case for `No Stored Identity exists for this user or Stored Identity has been invalidated` will now be covered by an OAuth error response from the new `/authorize` endpoint in the future.

openAPI/public-api.yaml
```diff
@@ -0,0 +1,118 @@
+---
+openapi: 3.0.3
+info:
+  title: "SIS Public API"
+  description: "Public API for the Stored Identity Service"
+  version: "1.0.0"
+paths:
+  /authorize:
+    get:
+      summary: Authorization Endpoint
```
michaelabenyohai: "Authorization Endpoint" sounds like a typo of "Authorize Endpoint", which is the OAuth terminology for this endpoint.

It would be nice to reword this to one of:

"OAuth2 Authorize Endpoint"

OR

"Get an authorization grant" (this is probably more consistent with the other plain English descriptions of the other endpoints)

src/handlers/user-identity-handler/user-identity-handler.ts
```diff
@@ -0,0 +1,11 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult, Context } from "aws-lambda";
+import logger from "../../commons/logger";
+
+export const handler = async (event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> => {
+  logger.addContext(context);
+
+  return {
+    statusCode: 200,
+    body: "{}",
```
michaelabenyohai: The story says to return a "dummy reponse". I wonder if it would be nice to put something in here that actually reflects the shape of the data that would be returned? Even if it's just random strings against some of the fields? That way we can see the data coming back when we start testing the integration of the new Orch stub with these endpoints?

openAPI/public-api.yaml
```diff
@@ -0,0 +1,118 @@
...
+        The authorization endpoint is used to interact with the resource owner
+        and obtain an authorization grant.
+      tags:
+        - Authorization
+      parameters:
+        - $ref: "#/components/parameters/ResponseTypeParam"
+        - $ref: "#/components/parameters/ClientIdParam"
+        - $ref: "#/components/parameters/RedirectUriParam"
+        - $ref: "#/components/parameters/ScopeParam"
+        - $ref: "#/components/parameters/StateParam"
```
michaelabenyohai, line 21: ℹ️  Not required for this PR, but for info I think all the data we need from Orchestration (like `vtr` and `sub`) will end up in a JAR which will in a query parameter called `request`. As per the [IPV Core frontend code](https://github.com/govuk-one-login/ipv-core-front/blob/10ff3a044564c7edd62efc60e6475f5e1a1d80d3/src/app/oauth2/middleware.ts#L17).

But we could always leave this out of scope for this PR and add it when we come to implementing the `/authorize` endpoint properly.

openAPI/public-api.yaml
```diff
@@ -0,0 +1,118 @@
...
+      responses:
+        "302":
+          $ref: "#/components/responses/AuthorizationRedirect"
+        "400":
```
michaelabenyohai: I'm not certain, but I'm fairly sure we wouldn't use a 400 response code. The [OAuth spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) says that we should use a `302` redirect with `invalid_request` error code, unless:

> the request fails due to a missing, invalid, or mismatching redirection URI, or if the client identifier is missing or invalid, the authorization server"

And our own RFC 0014 says that in the case of a problematic redirect_uri or JAR we should return a `500`.

I think it might be worth removing `400` and adding `500`? Or just remove `400` and we can revisit the error scenarios when they are implemented at a later date.

tests/acceptance/steps/utils/auth-api.ts
```diff
@@ -0,0 +1,96 @@
...
+export const AuthorizationResponseType = {
+  code: "code",
+} as const;
+
+export type AuthorizationParams = {
+  response_type: keyof typeof AuthorizationResponseType;
+  client_id: string;
+  redirect_uri: string;
+  scope?: string;
+  state?: string;
```
michaelabenyohai: As per other comments, can we make `state` not optional?

tests/acceptance/steps/utils/auth-api.ts
```diff
@@ -0,0 +1,96 @@
...
+  const publicApi = await getCloudFormationOutput(CloudFormationOutputs.SisPublicApi);
+
+  const response = await request(publicApi).get("/authorize").query(params).send();
+  if (isAwsError(response.body)) {
+    return { error: "error", error_description: response.body.message };
+  }
+
+  const code = /https:.*code=(.*)/.exec(response.header["location"]);
+  if (code) {
+    return { code: code[1] };
```
michaelabenyohai: If the `response.header["location"]` is `https://client.example.com/cb?code=SplxlOBeZQQYbYS6WxSbIA&state=xyz` (i.e. it includes a `state`) then I believe this code would return:

`{ code: "SplxlOBeZQQYbYS6WxSbIA&state=xyz" }`

but I think we want it to return:

`{ code: "SplxlOBeZQQYbYS6WxSbIA" }`

tests/acceptance/steps/authorization.step.ts
```diff
@@ -0,0 +1,64 @@
...
+When<WorldDefinition>("I call the authorize endpoint", async function () {
+  this.authorizationResponse = await authorize({
+    client_id: "acceptance-tests",
+    state: "acceptance-tests",
+    response_type: AuthorizationResponseType.code,
+    redirect_uri: "https://api.example.com",
+  });
+});
+
+Then<WorldDefinition>("I will be issued with an authorization code and redirected to the endpoint", function () {
```
michaelabenyohai: The test description implies that it is testing that the result is a redirection to the correct endpoint, but this doesn't actually seem to test where we are redirected to.

I think it would be nice to test that we are redirected to the correct endpoint - i.e. to the `redirect_uri` specified in the earlier `I call the authorize endpoint` step.

tests/acceptance/steps/authorization.step.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  this.tokenResponse = await token({
+    grant_type: TokenGrantType.AuthorizationCode,
+    code: this.authorizationResponse.code,
+  });
+});
+
+Then<WorldDefinition>("I will be issued with an authorization token", function () {
+  assert.ok(isTokenResponse(this.tokenResponse));
+  assert.ok(this.tokenResponse.access_token);
```
michaelabenyohai, line 50: It would be nice to assert that the `token_type` is `"Bearer"`

tests/acceptance/steps/utils/sis-api.ts
```diff
@@ -20,3 +20,17 @@ export async function sisPostUserIdentity(data: SisPostBody, bearerToken?: strin
 
   return await requestOperation;
 }
+
+export async function sisPostUserIdentityHandler(data: SisPostBody, bearerToken?: string): Promise<Response> {
+  const requestOperation = request(await getCloudFormationOutput(CloudFormationOutputs.SisPrivateApiAcceptanceTest))
+    .post("/user-identity")
+    .send(JSON.stringify(data))
```
michaelabenyohai: As per a [related comment](https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/265/changes#r3080386825) I don't believe this new `/user-identity` endpoint takes a request body at all, so the code forming the body/content-type can be removed.

tests/acceptance/steps/utils/sis-api.ts
```diff
@@ -20,3 +20,17 @@ export async function sisPostUserIdentity(data: SisPostBody, bearerToken?: strin
...
+    .set("Accept", "*/*")
+    .set("Content-Type", "application/json");
+
+  if (bearerToken) {
+    requestOperation.set("Authorization", bearerToken);
```
michaelabenyohai: I believe for bearer tokens the header needs to be, for example, `Authorization: Bearer mF_9.B5f-4.1JqM`. So something like:

```requestOperation.set("Authorization", `Bearer  ${bearerToken}`)```

as per https://datatracker.ietf.org/doc/html/rfc6750#section-2.1

File: tests/acceptance/steps/utils/auth-api.ts (hunk 23 above)
Comment (chris-dco) [reply tests/acceptance/steps/utils/auth-api.ts]: Updated.

File: openAPI/private-api.yaml (hunk 2 above)
Comment (michaelabenyohai) [reply openAPI/private-api.yaml]: Resolved

File: openAPI/public-api.yaml (hunk 3 above)
Comment (michaelabenyohai) [reply openAPI/public-api.yaml]: Resolved

File: tests/acceptance/steps/utils/auth-api.ts (hunk 24 above)
Comment (chris-dco) [reply tests/acceptance/steps/utils/auth-api.ts]: I have updated this to return the origin and code. I have also changed it to use the Javascript in built types to avoid this type of bug.

File: src/handlers/authorize-handler/tests/authorize-handler.test.ts
Comment (michaelabenyohai) [reply src/handlers/authorize-handler/tests/authorize-handler.test.ts:1]: Resolved

File: src/handlers/authorize-handler/authorize-handler.ts (hunk 7 above)
Comment (michaelabenyohai) [reply src/handlers/authorize-handler/authorize-handler.ts:10]: Resolved

File: src/handlers/authorize-handler/authorize-handler.ts (hunk 5 above)
Comment (michaelabenyohai) [reply src/handlers/authorize-handler/authorize-handler.ts]: Resolved 🚀

File: openAPI/public-api.yaml (hunk 22 above)
Comment (michaelabenyohai) [reply openAPI/public-api.yaml]: Now that we've removed the 400 code from the authorize-handler I think it would be good to remove this and add it back in if we decide in future that we need it.

File: src/handlers/authorize-handler/authorize-handler.ts (hunk 6 above)
Comment (michaelabenyohai) [reply src/handlers/authorize-handler/authorize-handler.ts]: Resolved 🚀

File: src/handlers/authorize-handler/authorize-handler.ts (hunk 8 above)
Comment (michaelabenyohai) [reply src/handlers/authorize-handler/authorize-handler.ts]: Resolved 🚀

File: tests/acceptance/steps/authorization.step.ts (hunk 25 above)
Comment (chris-dco) [reply tests/acceptance/steps/authorization.step.ts]: The `authorize` function now returns the endpoint and the code. I have updated the test to check both.

File: tests/acceptance/steps/utils/sis-api.ts (hunk 28 above)
Comment (chris-dco) [reply tests/acceptance/steps/utils/sis-api.ts]: I have updated this.

File: openAPI/public-api.yaml (hunk 19 above)
Comment (chris-dco) [reply openAPI/public-api.yaml]: Updated to "Get an authorization grant".

File: openAPI/private-api.yaml (hunk 15 above)
Comment (chris-dco) [reply openAPI/private-api.yaml]: Updated.

File: openAPI/private-api.yaml (hunk 16 above)
Comment (chris-dco) [reply openAPI/private-api.yaml]: Updated.

File: openAPI/private-api.yaml (hunk 14 above)
Comment (chris-dco) [reply openAPI/private-api.yaml]: Removed and removed object schema.

File: openAPI/private-api.yaml (hunk 17 above)
Comment (chris-dco) [reply openAPI/private-api.yaml]: Updated and removed `StoredIdentityResponse`.

File: openAPI/private-api.yaml (hunk 18 above)
Comment (chris-dco) [reply openAPI/private-api.yaml]: Removed.

File: openAPI/public-api.yaml (hunk 21 above)
Comment (chris-dco) [reply openAPI/public-api.yaml:21]: Leaving for now.

File: openAPI/public-api.yaml (hunk 22 above)
Comment (chris-dco) [reply openAPI/public-api.yaml]: Removed 400.

File: src/handlers/token-handler/token-handler.ts (hunk 9 above)
Comment (chris-dco) [reply src/handlers/token-handler/token-handler.ts]: Updated to `invalid_request`

File: src/handlers/token-handler/token-handler.ts (hunk 10 above)
Comment (chris-dco) [reply src/handlers/token-handler/token-handler.ts]: Updated to `invalid_grant`

File: src/handlers/token-handler/token-handler.ts (hunk 11 above)
Comment (chris-dco) [reply src/handlers/token-handler/token-handler.ts]: Updated to `invalid_grant`

File: src/handlers/token-handler/token-handler.ts (hunk 12 above)
Comment (chris-dco) [reply src/handlers/token-handler/token-handler.ts]: Removed.

File: src/handlers/token-handler/token-handler.ts (hunk 13 above)
Comment (chris-dco) [reply src/handlers/token-handler/token-handler.ts]: Updated to `bearer`

File: tests/acceptance/steps/utils/sis-api.ts (hunk 27 above)
Comment (chris-dco) [reply tests/acceptance/steps/utils/sis-api.ts]: Updated to a `get` and removed body.

File: tests/acceptance/steps/authorization.step.ts (hunk 26 above)
Comment (chris-dco) [reply tests/acceptance/steps/authorization.step.ts:50]: Updated to include this.

File: openAPI/public-api.yaml (hunk 22 above)
Comment (michaelabenyohai) [reply openAPI/public-api.yaml]: Resolved 🚀

File: openAPI/public-api.yaml [hunk 29]
```diff
@@ -0,0 +1,116 @@
...
+        Successful authorization. The server redirects the user-agent back to the
+        client's `redirect_uri` with the authorization `code` and the `state`.
+      headers:
+        Location:
+          description: The redirect URL containing the code and state as query parameters.
+          schema:
+            type: string
+            format: uri
+            example: https://orch.stubs.account.gov.uk/callback?code=SplxlOBeZQQYbYS6WxSbIA&state=xyz
+    OAuthBadRequest:
```
michaelabenyohai: This code for `OAuthBadRequest` is no longer used so it would be nice to remove it

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

src/handlers/token-handler/token-handler.ts
```diff
@@ -45,8 +45,7 @@ export const handler = async (event: APIGatewayProxyEvent, context: Context): Pr
     statusCode: 200,
     body: JSON.stringify({
       access_token: createHash("sha256").update(code).digest("hex"),
-      scope: "unknown",
-      token_type: "example",
+      token_type: "bearer",
```
michaelabenyohai: This needs to be a capital B `"Berarer"` as per https://datatracker.ietf.org/doc/html/rfc6750#section-4

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

openAPI/private-api.yaml
michaelabenyohai: Resolved 🚀

openAPI/private-api.yaml
michaelabenyohai: Resolved 🚀

openAPI/private-api.yaml
michaelabenyohai: Resolved 🚀

openAPI/private-api.yaml
michaelabenyohai: Resolved 🚀

openAPI/private-api.yaml
michaelabenyohai: Resolved 🚀

openAPI/public-api.yaml
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/auth-api.ts
```diff
@@ -0,0 +1,98 @@
...
+  scope: string;
+  token_type: string;
+  expires_in: number;
+};
+
+export type AwsError = {
+  message: string;
+};
+
+export const isOAuthBadRequest = (response: unknown): response is OAuthBadRequest => {
```
michaelabenyohai: This code isn't used. I wonder if it is worth removing? And maybe also the corresponding `OAuthBadRequest` type?

tests/acceptance/steps/utils/auth-api.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/auth-api.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/authorization.step.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/authorization.step.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/sis-api.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/sis-api.ts
```diff
@@ -20,3 +20,16 @@ export async function sisPostUserIdentity(data: SisPostBody, bearerToken?: strin
 
   return await requestOperation;
 }
+
+export async function sisGetUserIdentityHandler(data: SisPostBody, bearerToken?: string): Promise<Response> {
+  const requestOperation = request(await getCloudFormationOutput(CloudFormationOutputs.SisPrivateApiAcceptanceTest))
+    .get("/user-identity")
+    .set("Accept", "*/*")
+    .set("Content-Type", "application/json");
```
michaelabenyohai: I think we can remove the `Content-Type` header now that this is a `get`

tests/acceptance/steps/utils/sis-api.ts
michaelabenyohai: Resolved 🚀

openAPI/public-api.yaml
chris-dco: Removed.

src/handlers/token-handler/token-handler.ts
chris-dco: Updated

tests/acceptance/steps/utils/auth-api.ts
chris-dco: Removed

tests/acceptance/steps/utils/sis-api.ts
chris-dco: Removed

openAPI/public-api.yaml
michaelabenyohai: Resolved 🚀

src/handlers/token-handler/token-handler.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/auth-api.ts
michaelabenyohai: Resolved 🚀

tests/acceptance/steps/utils/sis-api.ts
michaelabenyohai: Resolved 🚀

michaelabenyohai:
LGTM 😍

mislam246 (approved):

PR #281, opened 2026-04-23
SPT-1903: Create Dummy FE Screen

mislam246:
### What changed
- Added dummy front end screen
- Added logic to route users from the `/authorize` to the `/confirm-details` endpoint
- Added logic to route users back to the `redirect-uri` passed in from the client, including the auth code and state in the query parameters

### Why did it change
This ticket is required as part of the initial skeleton work for Phase 3, specifically adding a dummy page to SIS front end to accompany the creation of the new SIS endpoints.

### Issue tracking
[SPT-1903](https://govukverify.atlassian.net/browse/SPT-1903)

## Testing
Tested on: https://dev-mi.reuse-identity.dev.account.gov.uk/authorize?response_type=bearer&state=state-id&client_id=client-id&redirect_uri=https://api.example.com

### Checklist for developers

- [x]  All commits are signed

### Checklist for reviewers

[SPT-1903]: https://govukverify.atlassian.net/browse/SPT-1903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


umaar (commented):
I noticed a lot of the feedback I left here: https://github.com/govuk-one-login/ipv-reuse-service-stubs/pull/214 also applies to this PR. Would you be ok with taking on those comments from that PR and applying it here without me having to copy and paste it across?

src/handlers/confirm-details/confirm-details-handler.ts
```diff
@@ -0,0 +1,43 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import nunjucks from "nunjucks";
+import confirmDetailsTemplate from "./index.njk";
```
umaar: See this comment: https://github.com/govuk-one-login/ipv-reuse-service-stubs/pull/214/changes#r3131166915 which would also apply to this PR (ignore the `express` stuff).

src/handlers/confirm-details/confirm-details-submission-handler.ts
```diff
@@ -0,0 +1,44 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import logger from "../../commons/logger";
+
+const extractEventData = async (eventBody: string[]) => {
```
umaar: See this comment: https://github.com/govuk-one-login/ipv-reuse-service-stubs/pull/214/changes#r3131168749 tl;dr use `URLSearchParams`. It will help simplify this code, and is built into Node.js, and will be less code for us to maintain/for something to go wrong

src/handlers/govuk-resources-uploader/govuk-resources-uploader.ts
```diff
@@ -0,0 +1,114 @@
+import {
```
umaar, line 1: @chris-dco is coming up with a solution to avoid the [duplication](https://github.com/govuk-one-login/ipv-reuse-service-stubs/blob/main/ais-producer-stub/src/govuk-resources-uploader/govuk-resources-uploader.ts).

_Edit_: Tech lead [decision](https://gds.slack.com/archives/C03EMG4LNGK/p1777025696559599?thread_ts=1776958259.135169&cid=C03EMG4LNGK) is to leave this as-is for now.

src/handlers/confirm-details/confirm-details-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+      body: nunjucks.renderString(confirmDetailsTemplateContents, {
+        assetPath: "./assets",
+        rootPath: ".",
+        code,
+        redirect_uri,
+        state,
+      }),
+      headers: {
+        "content-type": "text/html",
+        code: code,
```
umaar: Interesting what would be the reason for this? Just wanted to check it's correct and intentional? Who/what consumes `code` in the header like this?

Also `code: code` can be shortened to `code`.

src/handlers/confirm-details/confirm-details-submission-handler.ts
```diff
@@ -0,0 +1,44 @@
...
+  for (const value of eventBody) {
+    const key = value.substring(0, value.indexOf("="));
+    const start = value.indexOf("=") + 1;
+    obj[key] = value.slice(start);
+  }
+  return obj;
+};
+
+export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const valuesArray = event.body!.split("&");
```
umaar: Do check out [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) which can help here also. FYI if ever in JS you find yourself writing custom regular expressions, or doing lots of `indexOf`'s, there's normally (but not always) a more declarative approach available in vanilla JS. In this case, `URLSearchParams` should be able to alleviate most of this.

src/handlers/confirm-details/index.njk
```diff
@@ -0,0 +1,76 @@
...
+                                text: message
+                            }
+                        }
+                    ]
+                }) }}
+            </div>
+        {% else %}
+            <div class="govuk-grid-column-two-thirds">
+                <form id="submitForm" action="/confirm-details" method="post">
+                    <input type="hidden" id="redirectUri" name="redirectUri" value="{{redirect_uri}}""></input>
```
umaar: Looks like there's an extra quote `value="{{redirect_uri}}""` which ironically could exacerbate XSS implications.

src/handlers/confirm-details/index.njk
```diff
@@ -0,0 +1,76 @@
...
+                    <input type="hidden" id="code" name="code" value="{{code}}""></input>
```
umaar: [Void elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) should avoid end tags. This works though because HTML is forgiving, but we should follow the standard/spec

src/handlers/confirm-details/index.njk
```diff
@@ -0,0 +1,76 @@
...
+                            value: {
+                                text: message
+                            }
+                        }
+                    ]
+                }) }}
+            </div>
+        {% else %}
+            <div class="govuk-grid-column-two-thirds">
+                <form id="submitForm" action="/confirm-details" method="post">
```
umaar, line 55: Two points:

- Have we checked for proper escaping and making sure we're not vulnerable to XSS? e.g. `value="{{redirect_uri}}""`
- The [GDS Way](https://gds-way.digital.cabinet-office.gov.uk/manuals/security-overview-for-websites.html) security section includes recommendations (*must* rather than *should*) for what security headers need adding onto our website. Are we already using them and if not, I feel they should be added in where applicable.

src/handlers/confirm-details/index.njk
```diff
@@ -0,0 +1,76 @@
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/input/macro.njk" import govukInput %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
```
umaar: Maybe a silly question but are all these components being used? Noticed some are being imported but unused? It's better to only import things that are being used

src/handlers/confirm-details/index.njk
```diff
@@ -0,0 +1,76 @@
...
+{% from "govuk/components/error-summary/macro.njk" import govukErrorSummary %}
+{% from "govuk/components/panel/macro.njk" import govukPanel %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+
+{% extends "govuk/template.njk" %}
+{% block head %}
+    <link rel="stylesheet" href="{{ rootPath | default("", true) }}/govuk-frontend.min.css">
+    <base href="./confirm-details" />
```
umaar, line 10: How important is this, is it possible to avoid? This can have subtle, hard-to-debug effects on *all resources*. Totally understand if this can't be avoided, but thought to ask as I've had issues with the base tag in the past. Sometimes, absolute references, or references specified from the root level path, can be better. That might not be practical with our setup though?

No need to spend any real time on this one though. `<base>` is totally fair to use in HTML. It's only if there's an immediately obvious alternative technique that I'd suggest to remove the `<base>` tag.

template.yaml
```diff
@@ -997,6 +998,246 @@ Resources:
...
+        - Key: Name
+          Value: !Sub "/aws/lambda/${UploadGovUKFrontEndResourcesLambda}"
+
+  GovUKAssetsBucket:
+    # checkov:skip=CKV_AWS_18: "Ensure the S3 bucket has access logging enabled"
+    # checkov:skip=CKV_AWS_53: "Ensure S3 bucket has block public ACLs enabled"
+    # checkov:skip=CKV_AWS_54: "Ensure S3 bucket has block public policy enabled"
+    # checkov:skip=CKV_AWS_21: "Ensure the S3 bucket has versioning enabled"
+    # checkov:skip=CKV_AWS_55: "Ensure S3 bucket has ignore public ACLs enabled"
+    # checkov:skip=CKV_AWS_56: "Ensure S3 bucket has RestrictPublicBuckets enabled"
```
umaar: Just wanted to confirm this is all intentional? Are we potentially disabling valid checks, for example `Ensure S3 bucket has block public ACLs enabled`....if API Gateway is the resource accessing the bucket, then would such a check still be valid?

Disclaimer: I am not an AWS person so maybe someone else can advise also. My naive thought is that the bucket shouldn't be public since traffic is routed through API Gateway, and Gateway can access the bucket with the correct permission.

src/handlers/confirm-details/confirm-details-handler.ts
```diff
@@ -0,0 +1,43 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
```
umaar, line 1: I noticed `src/handlers/confirm-details/confirm-details-submission-handler.ts` _does_ have a test file here `src/handlers/confirm-details/tests/confirm-details-submission-handler.test.ts` but then `src/handlers/confirm-details/confirm-details-handler.ts` (this file) _does not_ have a test file. Should we add tests for this file?

src/handlers/confirm-details/confirm-details-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+const confirmDetailsTemplateContents = fs.readFileSync(confirmDetailsTemplate, "utf8");
+
+export type ConfirmDetailsQueryStringParameters = {
+  code: string;
+  redirect_uri: string;
+  state: string;
+};
+
+export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const { code, redirect_uri, state } = event.queryStringParameters as ConfirmDetailsQueryStringParameters;
```
umaar, line 16: Ignoring TypeScript, think we need a runtime check to check that these exist, otherwise they'd be `undefined` and would silently make their way into the template. Without having thought about this a lot, guess we'd also need to return an appropriate response in the case they don't exist?

src/handlers/confirm-details/confirm-details-handler.ts
```diff
@@ -0,0 +1,43 @@
...
+        "content-type": "text/html",
+        code: code,
+      },
+    };
+  } catch (err) {
+    logger.error(`Error in lambdaHandler event: ${err}`);
+    return {
+      statusCode: 500,
+      body: JSON.stringify({
+        message: err,
```
umaar: Just wondering if you got a chance to check what this actually returns. Because if `err` is an [Error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) object, I think this won't show the contents of that error message?

samconfig.toml
```diff
@@ -3,8 +3,8 @@ version = 0.1
 [default.deploy.parameters]
 resolve_s3 = true
 region = "eu-west-2"
-confirm_changeset = true
+confirm_changeset = false
```
umaar, line 6: Just confirming this is intentional?

src/handlers/confirm-details/tests/confirm-details-submission-handler.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const response = await lambdaHandler(event);
+  expect(response).toStrictEqual({
+    statusCode: 302,
+    body: "",
+    headers: {
+      Location: "https://api.example.com/?code=abc123&state=test-state-id",
+    },
+  });
+});
```
umaar, line 19: Maybe my eyes deceive me but I think this is identical to the test below

src/handlers/confirm-details/tests/confirm-details-submission-handler.test.ts
```diff
@@ -0,0 +1,45 @@
+import { APIGatewayEventRequestContextWithAuthorizer, APIGatewayProxyEvent } from "aws-lambda";
```
umaar, line 1: Currently, the test cases feel quite limited. I might have missed something but I think there are no negative tests. E.g. missing body, missing fields, etc. It might be good to add those in.

openAPI/public-api.yaml
```diff
@@ -32,6 +32,84 @@ paths:
...
+          Fn::GetAtt: ConfirmDetailsLambdaRestApiRole.Arn
+        type: aws_proxy
+    post:
+      parameters:
+        - in: query
+          name: redirect_uri
+          schema:
+            type: string
+      responses:
+        "200":
```
umaar: Is this correct? Good chance I'm getting confused but I think the tests say this should be a 302 possibly

openAPI/public-api.yaml
```diff
@@ -32,6 +32,84 @@ paths:
...
+          default:
+            statusCode: "200"
+            responseParameters:
+              method.response.header.Cache-Control: "'public, max-age=10'"
+              method.response.header.Content-Type: integration.response.header.Content-Type
+              method.response.header.Content-Length: integration.response.header.Content-Length
+              method.response.header.Timestamp: integration.response.header.Date
+        requestParameters:
+          integration.request.path.item: method.request.path.item
+        # passthroughBehavior: when_no_match
```
umaar: Do we need this commented out code?

lambda-layers/Makefile
```diff
@@ -0,0 +1,19 @@
...
+    $(if $(patsubst /,,$(2)),\
+        $(call find-up,$(1),$(patsubst %/,%,$(dir $(2)))),\
+    )\
+)
+
+PROJECT_ROOT := $(strip $(call find-up,$(TARGET_DIR),$(MAKEFILE_LIST)))
+
+build-GovUKFrontEndLayer:
+	@if [ ! -n "$(PROJECT_ROOT)" ]; then \
+		echo "The node_modules folder was no found."; \
```
umaar: Typo: `no found`

.github/workflows/build.yml
```diff
@@ -38,6 +38,10 @@ jobs:
           node-version-file: ".nvmrc"
           cache: npm
 
+      - name: Install Dependencies
+        working-directory: ./
```
umaar: Wouldn't this be the default already? Wondering if this is needed?

template.yaml
```diff
@@ -997,6 +998,246 @@ Resources:
...
+      BuildProperties:
+        EntryPoints:
+          - src/handlers/confirm-details/confirm-details-submission-handler.ts
+        Target: es2022
+        Minify: true
+        Sourcemap: true
+
+  ConfirmDetailsLambdaRestApiRole:
+    Type: AWS::IAM::Role
+    Condition: IsNotProduction
```
umaar, line 1179: An AWS person will have better insight here but if we're not wanting to create AWS resources in prod, then do the other resources also need `IsNotProduction`, right now some of the other resources do not have this condition

template.yaml
```diff
@@ -997,6 +998,246 @@ Resources:
...
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
+
+  ConfirmDetailsLambda:
```
umaar, line 1143: Treat this with a grain of salt, people with more experience in AWS will know more. But I was looking at the Authorization Lambda and noticed we might be missing some properties e.g. `Format: esm` / `Platform: node` and maybe more. Wonder if those are needed?

template.yaml
```diff
@@ -997,6 +998,246 @@ Resources:
...
+      Layers:
+        - !Ref GovUKFrontEndLayer
+    Metadata:
+      BuildMethod: esbuild
+      BuildProperties:
+        UseNpmCi: True
+        EntryPoints:
+          - src/handlers/govuk-resources-uploader/govuk-resources-uploader.ts
+        Sourcemap: true
+        Target: "es2020"
```
umaar: Sometimes we use `es2022`, best to be consistent.

umaar:
Nice work, looks really good! I left some comments that maybe @chrisclayson  could skim through also

openAPI/public-api.yaml
chrisclayson: Good catch ... this will always be a redirect

src/handlers/confirm-details/confirm-details-handler.ts
chrisclayson: Indeed ... we will be adding an error page, so any errors should be redirected to there ... but doesn't exist yet. I think, throw for the time being, and when we add the error page we can add the redirect.

src/handlers/confirm-details/confirm-details-handler.ts
chrisclayson: We don't need a `code` header.

src/handlers/confirm-details/confirm-details-handler.ts
chrisclayson: Good point. Also, we may want to avoid returning the raw content of an error (such as stack trace) in the response, we should definitely log it, but not return it.

chrisclayson: Though, as above, we're likely to replace this with a redirect to an error page.

template.yaml
chrisclayson: Correct ... I think `sam validate` should have picked this up as you have a conditional on the role but not the lambda.

_Really_ the only thing we don't want in production is the API Gateway (i.e. the front door that will let people through), I'm OK with the lambdas being deployed as they went do anything/cost us anything.

template.yaml
mislam246: @chrisclayson / @chris-dco do we need these additional properties on this Lambda?

template.yaml
mislam246: I'm not entirely sure. @chrisclayson / @chris-dco which of these do you think we should keep or remove?

src/handlers/confirm-details/confirm-details-submission-handler.ts
mislam246: That makes it much simpler, thanks for the suggestion. I've updated this to use `URLSearchParams` now.

src/handlers/confirm-details/confirm-details-handler.ts
mislam246: Initially had this for some local testing. Removed the header now.

src/handlers/confirm-details/confirm-details-submission-handler.ts
mislam246: Updated to use `URLSearchParams` as above.

src/handlers/confirm-details/index.njk
mislam246: Removed the extra quotes for all three inputs.

src/handlers/confirm-details/index.njk
mislam246: Good suggestion. I've removed the end tags for these inputs.

src/handlers/confirm-details/index.njk
mislam246: From what I've read, character escaping is enabled by default in Nunjucks. As we discussed, the testing of security headers is possibly a wider piece of work.

src/handlers/confirm-details/index.njk
mislam246: I've replaced the HTML inputs with the `govukInput` component and removed the `govukRadios` import. The rest of the components were being used originally.

src/handlers/confirm-details/index.njk
mislam246: I've faced problems getting it to work without the `<base>` tag. Given this will be revisited when we work on the front-end, is it okay to keep as is for now?

template.yaml
mislam246: As discussed, I've now removed all but the first line and added the `VersioningConfiguration` and `PublicAccessBlockConfiguration` properties to the bucket in the same way as this [example in SPOT](https://github.com/govuk-one-login/ipv-spot-key-management/blob/447b1d89cf9b4e2aef284d7b2002315104332b47/template.yaml#L974).

src/handlers/confirm-details/confirm-details-handler.ts
mislam246: Added a check on these parameters and thrown an error if any are undefined.

src/handlers/confirm-details/confirm-details-handler.ts
mislam246: It returns an object containing the error message. I've changed this to not return the error and only log it.

samconfig.toml
mislam246: I initially changed this to make dev deployments quicker for testing. If you feel it's safer, I'm happy to undo the change.

src/handlers/confirm-details/tests/confirm-details-submission-handler.test.ts
mislam246: That test is no longer needed, I've removed it.

openAPI/public-api.yaml
mislam246: Changed this response code to `302`.

openAPI/public-api.yaml
mislam246: Not required, I've removed it.

lambda-layers/Makefile
mislam246: Fixed the typo.

template.yaml
mislam246: Added the `IsNotProduction` condition on the other resources.

template.yaml
mislam246: As discussed, I've added the additional properties (`UseNpmCi`, `Format`, `Platform`, `Target`, `OutExtension`, and `Banner`) with the same values as this [SPOT function](https://github.com/govuk-one-login/ipv-spot/blob/a5663caced075c767606b1e1c1685c2c9dd68f61/template-spot-functions.yaml#L425).

template.yaml
mislam246: Changed this to use `es2022` instead.

src/handlers/confirm-details/confirm-details-handler.ts
mislam246: Added the `src/handlers/confirm-details/confirm-details-handler.test.ts` file with additional tests.

src/handlers/confirm-details/tests/confirm-details-submission-handler.test.ts
mislam246: I've added some tests to check an error is thrown when there are missing required URL query parameters in the request body.

.github/workflows/build.yml
mislam246: Tested without this and it's working 👍

src/handlers/confirm-details/confirm-details-handler.ts
umaar: @mislam246 I did some research into this and have fixed this in fa7109055cdff15be1466af90fc38f402a4f6328. We spoke yesterday about the _what_ so I won't repeat that here, but in case it's helpful, I wanted to explain the _why_, so it's clearer:

The problem with creating a custom `.njk=file` loader in esbuild: esbuild is a JavaScript/TypeScript bundler. Nunjucks templates are static text files that have nothing to do with JavaScript compilation. 

Creating a custom esbuild `.njk=file` loader to get a template into the Lambda artifact forces a chain of workarounds:

1. TypeScript module declaration (declare module `"*.njk"`): needed to make `import x from "./index.njk"` type-check, since TypeScript doesn't natively understand .njk imports.
2. Non-standard import pattern: `import template from "./index.njk"` only resolves to a file path because of the bespoke esbuild loader, not because of any standard JS/TS semantics.
3. `fs.readFileSync` + `nunjucks.renderString()`: a roundabout reimplementation of what `nunjucks.render()` already does with its built-in [FileSystemLoader](https://mozilla.github.io/nunjucks/api.html#filesystemloader).
4. We had to write a custom Vitest plugin (`nunjucks-raw-loader`) since tests can't resolve `.njk` imports either, so a custom Vite transform plugin was needed to make tests pass.
5. Content-hashed filenames: I believe esbuild's file loader outputs stuff like `index-abc123.njk` (content-hashed), so even the runtime filename is unpredictable.
6. ESM compatibility shim: when converting to ESM output, a Banner with `import { createRequire } from 'module'; const require = createRequire(import.meta.url);` was needed.

The proper way of doing this is to just copy the template directly. We then get a simpler API `nunjucksEnv.render("index.njk", context);`.

What this removes:
- `src/types.d.ts`: the declare module `"*.njk"` TypeScript declaration
- `.njk=file` esbuild loader config
- `fs.readFileSync` + `nunjucks.renderString()` pattern
- `nunjucks-raw-loader` custom vitest plugin and assetsInclude config
- The ESM Banner shim
- `import confirmDetailsTemplate from "./index.njk"`: no importing templates in JS

I've deployed to dev, and confirmed it's working. Happy to explain more if any questions

mislam246: Thanks for the fix and the detailed explanation 👍

.github/workflows/build.yml
chris-dco, line 56: Add this to fix your build error: 
```suggestion
            openAPI/
            lambda-layers/
```

mislam246: Thanks. For future reference, the missing GitHub permissions also needed adding to `gha-aws-oidc-provider-template` and then running the `./apply-role-changes-dev.sh` script.

src/handlers/confirm-details/confirm-details-handler.ts
umaar: A check has been added, but it's inconsistent: If state is missing from the query string, it would be `undefined`, which is not `=== ""`.  

Also, relying on [loose equality](https://eslint.org/docs/latest/rules/eqeqeq) to catch both null and undefined is a subtle JS quirk that's easy to misread or break. We should probably apply that eslint rule in the future.

For now, to fix this, you could just do `if (!code || !redirect_uri || !state) {` which also catches empty strings. Maybe worth thinking about whether or not the unit tests would've caught that - no harm in adding a test in for that.

samconfig.toml
umaar: Oh should be ok I think, was just to check there was a reason

src/handlers/confirm-details/confirm-details-handler.ts
mislam246: Makes sense, thanks for clarifying. I've refactored it as suggested and added another test to check for empty strings.

umaar (dismissed):

chrisclayson (approved):

PR #304, opened 2026-05-11
BAU: Added retry logic for consumer tests

chris-dco:
### What changed

Added retry logic for consumer tests

### Why did it change

This is due to communications issues that we have observed between the RUST PACT server and the PACT client. It appears that the PACT server handles the request, however the PACT client does not receive the expected response. We have compared the logs between a working test run and a failed test run and there are identical leading us to assume that there are comms issues.

### DIFF
--- a/src/tests/contract-tests/vitest.config.ts
+++ b/src/tests/contract-tests/vitest.config.ts
@@ -4,5 +4,10 @@ export default defineConfig({
   test: {
     projects: ["src/tests/contract-tests/consumer", "src/tests/contract-tests/provider"],
     bail: 1,
+    retry: {
+      // Retry to get around comms issues between PACT RUST server and JS client
+      count: 2,
+      delay: 50,
+    },
   },
 });


src/tests/contract-tests/vitest.config.ts
```diff
@@ -4,5 +4,10 @@ export default defineConfig({
   test: {
     projects: ["src/tests/contract-tests/consumer", "src/tests/contract-tests/provider"],
     bail: 1,
+    retry: {
+      // Retry to get around comms issues between PACT RUST server and JS client
```
umaar, line 8: Wondering if you feel it's practical to identify (and maybe fix) the root cause? Simply retrying tests might mask what the actual issue is, and simply push the problem down the line. My input here is limited as I don't understand the contract tests, but I know for acceptance tests, I would be recommending we fix the root cause.

chris-dco: Essentially it's a problem inside of a black box we have no control over.

umaar: I struggle to understand anything with this comment. Where does the problem lie, does a bug need raising on a PACT repo? Is it something to do with our infrastructure? Are all tests affected? Are intermittent failures always fixed on the 2nd try?

PR #319, opened 2026-06-01
SPT-1948: Adds OpenAPI spec

umaar:
### What changed

- Create openApi spec for OAuth common: 3 endpoints (POST /session, GET /authorization and POST /token endpoints) at `openAPI/oauth-common-private-api.yaml`

### Why did it change

- Prep work for future tasks which will wire up the lambdas 

### Issue tracking

- SPT-1948

### Manual Steps to Test

- Compare with the spec from HMRC check CRI


openAPI/oauth-common-private-api.yaml
```diff
@@ -0,0 +1,215 @@
...
+        redirect_uri:
+          type: "string"
+          format: "uri"
+          example: "https://orch.stubs.account.gov.uk/callback"
+      additionalProperties: false
+
+    TokenResponse:
+      required:
+        - "access_token"
+        - "token_type"
```
joe-mcgarry: I could very well have missed something, so wanted to check but is `token_type` meant to be required? It doesn't appear to be in the HMRC response but I might be forgetting something!

umaar: Fair point! Have removed it. I noticed our existing `private-api.yaml` did mark it as required, so had initially thought to align it with that. But maybe HMRC parity is more important!

Interestingly it's also required according to the [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1).

joe-mcgarry: If it's required, then fair play! I was just curious as all I could see on the ticket was a reference to how the HMRC CRIs work, I'll leave it up to you (unless someone chimes in) as to whether it's required or not!

umaar: I don't know for sure! But it's a good catch. Maybe @chrisclayson has some thoughts. Otherwise, maybe aligning with HMRC is the safer option for now.

chrisclayson: So this must exactly match the HMRC one linked in the ticket. The reason being this is not the exposed OAuth2 endpoints themselves, more this is an internal "helper" API for OAuthy operations, as such it does not match the OAuth [spec](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) (though there are similarities).

This will be called by the `/token` endpoint on the API that our clients call (which does conform to spec).

umaar: Got it, thank you! Currently, this PR should now match with HMRC's version

joe-mcgarry: Going to add another question here, on [L349](https://github.com/govuk-one-login/ipv-cri-check-hmrc-api/blob/496f781bbd120ca9c31e09aec89a161f353ee4d0/infrastructure/public-api.yaml#L349) in the HMRC TokenResponse, they've got `refresh_token` which isn't applicable. 

Does that mean it's fine to not be part of our TokenResponse (as far as I can see, please do correct me if I've missed it) or would it still be needed? 

Sorry to be a pain!

umaar: All good! It really isn't being a pain and these are good questions that helps me (and possibly others) understand. Thank you.

If you look a few lines down that link you gave, exactly as you've pointed out, it says `description: "The refresh token is optional and not currently applicable"`, and SIS wouldn't use it at all.

Rather than copying something that was unused, I thought better to leave it out. Happy to add it back in if you like. My personal suggestion is that since there's no need for it, we don't need to copy it across.

joe-mcgarry: Completely get you, my brain was thinking the exact same thing when I was reading the HMRC spec! 

My thoughts are very much like yours: if we know SIS won't need it, and we know it isn't applicable within the spec we're aligning with, why add it if it's not needed

joe-mcgarry (approved):

PR #340, opened 2026-06-12
SPT-1916: Create error page

mislam246:
### What changed
- Created a new front end error page in SIS
- Renamed existing AWS resources to be shared between the front end related components

### Why did it change
As part of the implementation of a front-end for SIS Phase 3, an error page is required to route a user to when they encounter a problem.

### Issue tracking
[SPT-1916](https://govukverify.atlassian.net/browse/SPT-1916)

## Testing
Tested in the dev environment: https://preview-pr-340.reuse-identity.dev.account.gov.uk/error/unrecoverable

### Checklist for developers

- [x]  All commits are signed

### Checklist for reviewers

[SPT-1916]: https://govukverify.atlassian.net/browse/SPT-1916?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,3 +8,14 @@ build-ConfirmDetailsLambda:
 		--external:govuk-frontend \
 		--outdir=$(ARTIFACTS_DIR)
 	cp src/handlers/confirm-details/*.njk $(ARTIFACTS_DIR)/
+
+build-UnrecoverableErrorLambda:
+	./node_modules/.bin/esbuild src/handlers/unrecoverable-error/unrecoverable-error-handler.ts \
+		--bundle \
+		--platform=node \
+		--target=es2022 \
+		--minify \
+		--sourcemap \
+		--external:govuk-frontend \
+		--outdir=$(ARTIFACTS_DIR)
+	cp src/handlers/unrecoverable-error/*.njk $(ARTIFACTS_DIR)/
--- a/src/handlers/unrecoverable-error/unrecoverable-error-handler.ts
+++ b/src/handlers/unrecoverable-error/unrecoverable-error-handler.ts
@@ -1,0 +1,28 @@
+import { APIGatewayProxyResult } from "aws-lambda";
+import nunjucks from "nunjucks";
+import path from "node:path";
+import logger from "../../commons/logger";
+
+const govukFrontendDistribution = path.join(path.dirname(require.resolve("govuk-frontend/package.json")), "dist");
+const nunjucksEnvironment = nunjucks.configure([__dirname, govukFrontendDistribution]);
+
+export const lambdaHandler = async (): Promise<APIGatewayProxyResult> => {
+  try {
+    return {
+      statusCode: 200,
+      body: nunjucksEnvironment.render("index.njk", {
+        assetPath: "/assets",
+        rootPath: "",
+      }),
+      headers: {
+        "content-type": "text/html",
+      },
+    };
+  } catch (error) {
+    logger.error(`Error in lambdaHandler event: ${error}`);
+    return {
+      statusCode: 500,
+      body: "",
+    };
+  }
+};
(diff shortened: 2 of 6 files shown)


src/handlers/unrecoverable-error/unrecoverable-error-handler.ts
```diff
@@ -0,0 +1,29 @@
...
+import nunjucks from "nunjucks";
+import path from "node:path";
+import logger from "../../commons/logger";
+
+const govukFrontendDistribution = path.join(path.dirname(require.resolve("govuk-frontend/package.json")), "dist");
+const nunjucksEnvironment = nunjucks.configure([__dirname, govukFrontendDistribution]);
+
+export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  try {
+    const stage = event.requestContext?.stage ? `/${event.requestContext.stage}` : "";
```
chris-dco: We don't use stages for the pretty URL. How come you have it here?

template.yaml
```diff
@@ -1151,7 +1151,7 @@ Resources:
     Properties:
       FunctionName: !Sub "${AWS::StackName}-ConfirmDetailsLambda"
       Handler: confirm-details-handler.lambdaHandler
-      Role: !GetAtt ConfirmDetailsRole.Arn
+      Role: !GetAtt FrontEndAssetsAccessRole.Arn
```
chris-dco: I would give each of the Lambdas their own separate roles. This is so that we ensure we're following the principle of least privilege and we do not accidentally give a Lambda roles in the future which it does not need.

Makefile
```diff
@@ -8,3 +8,14 @@ build-ConfirmDetailsLambda:
...
+build-UnrecoverableErrorLambda:
+	./node_modules/.bin/esbuild src/handlers/unrecoverable-error/unrecoverable-error-handler.ts \
+		--bundle \
+		--platform=node \
+		--target=es2022 \
+		--minify \
+		--sourcemap \
+		--external:govuk-frontend \
+		--outdir=$(ARTIFACTS_DIR)
+	cp src/handlers/unrecoverable-error/*.njk $(ARTIFACTS_DIR)/
```
chrisclayson, line 21: <del>I think we're at the stage where we need to make some decision around this ... I think this `cp` statement runs the risk of over-writing assets copied in on line `10` which may, or may not, be the same versions 🤔</del>

chrisclayson: Ignore, I was confused, I thought this was the Makefile for the lambda layer.

src/handlers/unrecoverable-error/unrecoverable-error-handler.ts
mislam246: Due to the new API endpoint having two path segments (`/error/unrecoverable`), it was using `/error` as the resource path, therefore looking for the front-end assets in the wrong directory and not loading the CSS. Also, in the dev environment it has the `/v1` stage. So this change would set the resource path to the correct root directory of the front-end assets, even for the dev environment with its stage name.

template.yaml
mislam246: Makes sense, I initially made it a shared role to reduce duplication but happy to change. I've separated it out now so the new Lambda has its own roles.

src/handlers/unrecoverable-error/unrecoverable-error-handler.ts
chris-dco: We don't use stages and v1 never appears in the URL path. For example, the URL for this PR is https://preview-pr-340.reuse-identity.dev.account.gov.uk/error/unrecoverable

mislam246: I've removed the `stage` logic and updated the `assetPath` and `rootPath` which now correctly specifies the resources path and applies the CSS.

chris-dco (approved):

PR #363, opened 2026-06-25
SPT-1969 - Rename SIS Functions to be more distinctive

chris-dco:
## What

* [x] Renamed `AuthorizationFunction` to `GetAuthorizeHandler`
* [x] Renamed `TokenFunction` to `PostTokenHandler`
* [x] Renamed `UserIdentityHandlerFunction` to `GetUserIdentityHandler`
* [x] Renamed `ConfirmDetailsLambda` to `GetConfirmDetailsHandler`
* [x] Renamed `ConfirmDetailsSubmissionLambda` to `PostConfirmDetailsHandler`
* [x] Renamed `UnrecoverableErrorLambda` to `GetUnrecoverableErrorHandler`
* [ ] Renamed `UserIdentityFunction` to PostPhase2UserIdentityHandler`
* [ ] Renamed `MessageProcessorLambda` to InboundTxmaQueueMessageHandler`

> [!NOTE]
> Renaming `UserIdentityFunction` and `MessageProcessorLambda` will not be done in this PR and will be done separately. This is because those functions are in production and the change needs to be made in stages in order to guarantee it will be zero-downtime. For the rest of the lambdas (in this PR) they are only depoyed up to staging, so we do not care about the deployment being zero downtime.

## Why

In order to enable SIS to make use of Common OAuth components, SIS and it’s Orchestration stub will require updates to integrate into the OAuth flow. To avoid further confusion, the SIS functions require renaming so that the names are much clearer and recognisable as to their purpose.


template.yaml
```diff
@@ -350,14 +350,14 @@ Resources:
   # AUTHORIZATION LAMBDA         #
```
michaelabenyohai: Maybe worth renaming the section (and any others)?

template.yaml
```diff
@@ -532,14 +532,14 @@ Resources:
...
```
michaelabenyohai: Looks like this was always wrong, can you update to be the new name for token?

template.yaml
```diff
@@ -350,14 +350,14 @@ Resources:
...
   ################################
 
-  AuthorizationFunction:
+  GetAuthorizeHandler:
     Type: AWS::Serverless::Function
     Condition: IsNotProduction
     Properties:
-      FunctionName: !Sub "${AWS::StackName}-AuthorizationFunction"
+      FunctionName: !Sub "${AWS::StackName}-GetAuthorizeHandler"
       Description: "A lambda function that performs OAuth"
```
michaelabenyohai: Whilst we're here, can we reword the `Description` to be:

`A lambda function that handles a request to the OAuth 2.0 authorization server's authorization endpoint`

template.yaml
```diff
@@ -532,14 +532,14 @@ Resources:
...
   ################################
 
-  TokenFunction:
+  PostTokenHandler:
     Type: AWS::Serverless::Function
     Condition: IsNotProduction
     Properties:
-      FunctionName: !Sub "${AWS::StackName}-TokenFunction"
+      FunctionName: !Sub "${AWS::StackName}-PostTokenHandler"
       Description: "A lambda function that performs OAuth"
```
michaelabenyohai: Whilst we're here, can we reword the Description to be:

`A lambda function that handles a request to the OAuth 2.0 authorization server's token endpoint`

template.yaml
```diff
@@ -741,13 +741,13 @@ Resources:
   ################################
   # POST USER IDENTITY FEATURE  #
```
michaelabenyohai: Let's change the section to "GET"

template.yaml
```diff
@@ -1136,34 +1136,13 @@ Resources:
...
-            Principal:
-              Service: lambda.amazonaws.com
-            Action:
-              - "sts:AssumeRole"
-      ManagedPolicyArns:
-        - arn:aws:iam::aws:policy/AmazonS3ReadOnlyAccess
-        - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
-
-  ConfirmDetailsLambda:
+  GetConfirmDetailsLambda:
```
michaelabenyohai: I think this should be `GetConfirmDetailsHandler` everywhere (including all the roles and policies), instead of `GetConfirmDetailsLambda`

template.yaml
```diff
@@ -1240,28 +1220,27 @@ Resources:
...
             Resource:
-              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-ConfirmDetailsLambda:*"
-              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-ConfirmDetailsSubmissionLambda:*"
+              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-GetConfirmDetailsLambda:*"
       Roles:
-        - !Ref ConfirmDetailsLambdaRestApiRole
+        - !Ref GetConfirmDetailsLambdaRestApiRole
 
-  ConfirmDetailsFunctionLogGroup:
+  GetConfirmDetailsFunctionLogGroup:
```
michaelabenyohai: Should be `GetConfirmDetailsHandlerLogGroup`

michaelabenyohai:
Are any of the changes not zero-downtime given the nature of canary deploys? I'm thinking about the rename of things in production - so the `PostPhase2UserIdentityHandler` and `InboundTxmaQueueMessageHandler`?

chris-dco:
> Are any of the changes not zero-downtime given the nature of canary deploys? I'm thinking about the rename of things in production - so the `PostPhase2UserIdentityHandler` and `InboundTxmaQueueMessageHandler`?

All changes should deploy and then only after the API Gateway has been updated will the old resources be deleted. All changes should create new resources and should not remove permissions for existing resource.

template.yaml
mislam246: Renamed this section to `GET AUTHORIZE HANDLER` and others relevant to this ticket.

template.yaml
mislam246: Renamed to `POST TOKEN HANDLER`.

template.yaml
mislam246: Reworded the description as suggested.

template.yaml
mislam246: Updated the description.

template.yaml
mislam246: Changed to `GET`.

template.yaml
mislam246: Updated all instances to `GetConfirmDetailsHandler`.

template.yaml
mislam246: Renamed.

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -746,15 +746,15 @@ Resources:
       SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PrivateApiDev}/${ApiStageName}/POST/*"
 
   ################################
-  # POST USER IDENTITY FEATURE  #
+  # GET USER IDENTITY HANDLER    #
   ################################
-  UserIdentityHandlerFunction:
+  GetUserIdentityHandler:
```
mislam246, line 753: Would this change update a Production resource?

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -499,26 +499,26 @@ Resources:
...
               - "lambda:InvokeFunction"
             Resource:
-              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-AuthorizationFunction:*"
+              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-GetAuthorizeHandler:*"
       Roles:
-        - !Ref AuthorizationFunctionRestApiRole
+        - !Ref GetAuthorizeHandlerRestApiRole
 
-  AuthorizationFunctionLogGroup:
+  GetAuthorizeHandlerLogGroup:
```
mislam246, line 518: Do we want to add the `IsNotProduction` condition on this resource? It's probably inferred by CloudFormation, given the other resources have the condition, but wanted to double-check.

mislam246 (commented):
Added a couple of questions below:

infrastructure/identity-reuse-service/template.yaml
michaelabenyohai: Good catch - this function shouldn't be being used in production, yet it appears to be being deployed there. 

The `UserIdentityHandlerFunction` is the new dummy one that should only be deployed to staging. I might raise a separate PR to stop it being deployed in integration and prod first.

The `UserIdentityFunction` is the one that should be in prod (and not renamed in this PR) and

infrastructure/identity-reuse-service/template.yaml
michaelabenyohai: This has now been done and merged in another PR :)

infrastructure/identity-reuse-service/template.yaml
michaelabenyohai: This is not longer deployed in production, it was fixed in an earlier PR and this one has now been rebased on main.

mislam246 (approved):

PR #364, opened 2026-06-26
SPT-1973: Create /oauth/callback endpoint

mislam246:
#### [SPT-1973](https://govukverify.atlassian.net/browse/SPT-1973)
- Created a new `/oauth/callback` endpoint that calls the `/api/authorization` endpoint to retrieve the authorization code and redirect users on a successful request
- Included error handling logic for failed API requests and missing auth codes

**Note**: This change uses a test Session ID, which will be replaced with the actual cookie value once [SPT-1937](https://govukverify.atlassian.net/browse/SPT-1937) changes are applied

## Why
This endpoint will support the auth code exchange with Orchestration, and a user will be redirected to this endpoint after the auth code has been stored in the session table.

## Testing
Tested successful and unsuccessful scenarios in dev. Used the Orchestration stub to generate the JAR and manually called the `/api/session` endpoint with the JAR to create and store a Session ID in Dynamo. Using this Session ID value, I tested the `/oauth/callback` endpoint, which is returning the expected responses in each scenario.

[SPT-1973]: https://govukverify.atlassian.net/browse/SPT-1973?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SPT-1937]: https://govukverify.atlassian.net/browse/SPT-1937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openAPI/public-api.yaml
+++ b/openAPI/public-api.yaml
@@ -160,6 +160,18 @@ paths:
         requestParameters:
           integration.request.path.item: method.request.path.item
         type: aws
+  /oauth2/callback:
+    get:
+      responses:
+        "302":
+          $ref: "#/components/responses/AuthorizationRedirect"
+      x-amazon-apigateway-integration:
+        httpMethod: POST
+        uri:
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${GetCallbackHandler.Arn}:live/invocations
+        credentials:
+          Fn::GetAtt: GetCallbackHandlerRestApiRole.Arn
+        type: aws_proxy
 
 components:
   parameters:
--- a/src/handlers/get-callback-handler/get-callback-handler.ts
+++ b/src/handlers/get-callback-handler/get-callback-handler.ts
@@ -1,0 +1,86 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import logger from "../../commons/logger";
+import { URL } from "node:url";
+import { getCookieValues } from "../../commons/cookie-utilities";
+import { AuthorizationSuccessResponse, isValidSuccessResponse } from "./get-callback-response";
+import { isValidQueryParameters } from "./get-callback-request";
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const queryParameters = event.queryStringParameters || {};
+
+  const domainName = process.env.DOMAIN_NAME || "";
+
+  if (!isValidQueryParameters(queryParameters)) {
+    logger.error("Missing mandatory query string parameters to the GetCallbackHandler");
+    return await redirectToErrorPage(domainName);
+  }
+
+  const sessionId = getCookieValues(event)?.get("identity_reuse_service_session");
+
+  if (!sessionId) {
+    return await redirectToErrorPage(domainName);
+  }
+
+  const oauthInternalApiUrl = process.env.OAUTH_INTERNAL_API;
+  const url = new URL(`${oauthInternalApiUrl}/api/authorization`);
+
+  url.searchParams.append("client_id", queryParameters.client_id);
+  url.searchParams.append("redirect_uri", queryParameters.redirect_uri);
+  url.searchParams.append("response_type", "code");
+
+  try {
+    const responseFromAuthorizeEndpoint = await fetch(url, {
+      method: "GET",
+      headers: {
+        "session-id": sessionId,
+      },
+    });
+
+    if (responseFromAuthorizeEndpoint.status === 200) {
+      return await redirectToClient(responseFromAuthorizeEndpoint, domainName);
+    } else if (responseFromAuthorizeEndpoint.status === 403) {
+      return await redirectToClientWithError(queryParameters.redirect_uri, queryParameters.state);
+    } else {
+      logger.error(`${responseFromAuthorizeEndpoint.status} response code returned from the authorization endpoint`);
+      return await redirectToErrorPage(domainName);
+    }
+  } catch (error) {
+    logger.error(`Error in OAuth Callback handler event: ${error}`);
+    return await redirectToErrorPage(domainName);
+  }
+};
+
+const redirectToClient = async (authorizationResponse: Response, domainName: string) => {
+  if (!isValidSuccessResponse(await authorizationResponse.json())) {
+    logger.error("Invalid response properties received from authorization endpoint");
+    return await redirectToErrorPage(domainName);
+  }
+
+  const authorizationData = (await authorizationResponse.json()) as AuthorizationSuccessResponse;
+  const orchestrationRedirectUrl = new URL(decodeURIComponent(authorizationData.redirectionURI));
+  orchestrationRedirectUrl.searchParams.append("code", authorizationData.authorizationCode.value);
+  orchestrationRedirectUrl.searchParams.append("state", authorizationData.state.value);
+  return redirect(`${orchestrationRedirectUrl}`);
+};
+
+const redirectToClientWithError = async (redirectUri: string, state: string) => {
+  const orchestrationRedirectUrl = new URL(redirectUri);
+  orchestrationRedirectUrl.searchParams.append("error", "access_denied");
+  // To be replaced with the state returned by the /authorization endpoint once it's added to the oauth-common response
+  orchestrationRedirectUrl.searchParams.append("state", state);
+  return redirect(`${orchestrationRedirectUrl}`);
+};
+
+const redirectToErrorPage = async (domainName: string) => {
+  return await redirect(`https://${domainName}/error/unrecoverable`);
+};
+
+const redirect = async (location: string) => {
+  return {
+    statusCode: 302,
+    headers: {
+      Location: location,
+    },
+    body: "",
+  };
+};
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -1435,6 +1435,129 @@ Resources:
         - Key: Name
           Value: !Sub "/aws/lambda/${UnrecoverableErrorLambda}"
 
+  GetCallbackHandler:
+    Type: AWS::Serverless::Function
+    Condition: IsNotProduction
+    Properties:
+      FunctionName: !Sub "${AWS::StackName}-GetCallbackHandler"
+      Handler: get-callback-handler.handler
+      Role: !GetAtt GetCallbackHandlerRole.Arn
+      Environment:
+        Variables:
+          NODE_OPTIONS: "--enable-source-maps"
+          DOMAIN_NAME: !Ref ReuseIdentityCustomDomain
+          OAUTH_INTERNAL_API:
+            Fn::Sub:
+              - "${OauthInternalStackId}"
+              - OauthInternalStackId:
+                  Fn::ImportValue: !Sub "${OauthInternalStackName}-PrivateApi"
+      Layers:
+        - !Ref GovUKFrontEndLayer
+        - !If
+          - DynaTraceEnabled
+          - !Ref DynatraceNodeLayerArn
+          - !Ref AWS::NoValue
+    Metadata:
+      BuildMethod: esbuild
+      BuildProperties:
+        UseNpmCi: True
+        Minify: true
+        Target: es2022
+        Sourcemap: true
+        SourcesContent: false
+        EntryPoints:
+          - src/handlers/get-callback-handler/get-callback-handler.ts
+        Format: esm
+        Platform: node
+        OutExtension:
+          - .js=.mjs
+        Banner:
+          - js=import { createRequire } from 'module'; const require = createRequire(import.meta.url);
+
+  GetCallbackHandlerRole:
+    # checkov:skip=CKV_AWS_111: Ensure IAM policies does not allow write access without constraints
+    Type: AWS::IAM::Role
+    Condition: IsNotProduction
+    Properties:
+      PermissionsBoundary: !If
+        - UsePermissionsBoundary
+        - !Ref PermissionsBoundary
+        - !Ref AWS::NoValue
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
+
+  GetCallbackHandlerRestApiRole:
+    Type: AWS::IAM::Role
+    Condition: IsNotProduction
+    Properties:
+      RoleName: !Sub "${AWS::StackName}-GetCallbackHandlerRestApiRole"
+      Description: !Sub "Roles for the GetCallbackHandler"
+      PermissionsBoundary: !If [ UsePermissionsBoundary, !Ref PermissionsBoundary, !Ref AWS::NoValue ]
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: apigateway.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
+      Tags:
+        - Key: Product
+          Value: !Ref ProductTagValue
+        - Key: System
+          Value: !Ref SystemTagValue
+        - Key: Environment
+          Value: !Ref Environment
+        - Key: Owner
+          Value: !Ref OwnerTagValue
+        - Key: Source
+          Value: !Ref SourceTagValue
+
+  GetCallbackHandlerRestApiPolicy:
+    Type: AWS::IAM::Policy
+    Condition: IsNotProduction
+    Properties:
+      PolicyName: !Sub "${AWS::StackName}-GetCallbackHandlerRestApiPolicy"
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: "Allow"
+            Action:
+              - "lambda:InvokeFunction"
+            Resource:
+              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-GetCallbackHandler:*"
+      Roles:
+        - !Ref GetCallbackHandlerRestApiRole
+
+  GetCallbackHandlerLogGroup:
+    Type: AWS::Logs::LogGroup
+    Condition: IsNotProduction
+    Properties:
+      LogGroupName: !Sub "/aws/lambda/${GetCallbackHandler}"
+      RetentionInDays: 30
+      KmsKeyId: !GetAtt CloudWatchEncryptionKey.Arn
+      Tags:
+        - Key: Product
+          Value: !Ref ProductTagValue
+        - Key: System
+          Value: !Ref SystemTagValue
+        - Key: Environment
+          Value: !Ref Environment
+        - Key: Owner
+          Value: !Ref OwnerTagValue
+        - Key: Source
+          Value: !Ref SourceTagValue
+        - Key: Name
+          Value: !Sub "/aws/lambda/${GetCallbackHandler}"
+
   ################################
   # PRIVATE API API GATEWAY       #
   ################################
--- a/package-lock.json
+++ b/package-lock.json
@@ -5145,16 +5145,6 @@
         "node": "^20.17.0 || >=22.9.0"
       }
     },
-    "node_modules/hosted-git-info/node_modules/lru-cache": {
-      "version": "11.2.4",
-      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.2.4.tgz",
-      "integrity": "sha512-B5Y16Jr9LB9dHVkh6ZevG+vAbOsNOYCX+sXvFWFu7B3Iz5mijW3zdbMyhsh8ANd2mSWBYdJgnqi+mL7/LrOPYg==",
-      "dev": true,
-      "license": "BlueOak-1.0.0",
-      "engines": {
-        "node": "20 || >=22"
-      }
-    },
     "node_modules/html-escaper": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/html-escaper/-/html-escaper-2.0.2.tgz",
@@ -5936,6 +5926,16 @@
       "integrity": "sha512-HDWXG8isMntAyRF5vZ7xKuEvOhT4AhlRt/3czTSjvGUxjYCBVRQY48ViDHyfYz9VIoBkW4TMGQNapx+l3RUwdA==",
       "dev": true,
       "license": "MIT"
+    },
+    "node_modules/lru-cache": {
+      "version": "11.5.1",
+      "resolved": "https://registry.npmjs.org/lru-cache/-/lru-cache-11.5.1.tgz",
+      "integrity": "sha512-RPimw/7aMdv2oqRrxKwvZXcPfwBrn/JZ2xYcY9Hus/6LaS3VOAKVWKWgNLCFSiOm1ESXinjsDlidVU7JlnCN2A==",
+      "dev": true,
+      "license": "BlueOak-1.0.0",
+      "engines": {
+        "node": "20 || >=22"
+      }
     },
     "node_modules/luxon": {
       "version": "3.7.2",
--- a/src/commons/cookie-utilities.ts
+++ b/src/commons/cookie-utilities.ts
@@ -1,0 +1,33 @@
+import { APIGatewayProxyEvent } from "aws-lambda";
+import logger from "./logger";
+
+export const getCookieValues = (event: Partial<APIGatewayProxyEvent>): Map<string, string> | undefined => {
+  const lowercaseHeaders = Object.fromEntries(
+    Object.entries(event.headers!).map(([key, value]) => [key.toLowerCase(), value])
+  );
+
+  const cookieHeader = lowercaseHeaders["cookie"];
+
+  if (!cookieHeader) {
+    return undefined;
+  }
+
+  const cookies = cookieHeader.split(";");
+
+  const cookieValues = new Map<string, string>();
+  for (const element of cookies) {
+    const cookie = element.trim();
+    const separatorIndex = cookie.indexOf("=");
+    const cookieName = cookie.slice(0, separatorIndex);
+    const cookieValue = cookie.slice(separatorIndex + 1);
+    try {
+      const decodedCookieValue = decodeURIComponent(cookieValue!);
+      cookieValues.set(cookieName, decodedCookieValue);
+    } catch (error) {
+      logger.error(`Error decoding cookie: ${error}`);
+      return undefined;
+    }
+  }
+
+  return cookieValues;
+};
--- a/src/commons/tests/cookie-utilities.test.ts
+++ b/src/commons/tests/cookie-utilities.test.ts
@@ -1,0 +1,49 @@
+import { describe, expect, it, vitest } from "vitest";
+import { APIGatewayEventRequestContextWithAuthorizer, APIGatewayProxyEvent } from "aws-lambda";
+import { getCookieValues } from "../cookie-utilities";
+
+describe("cookie-utilities", () => {
+  it("should return all cookie values", async () => {
+    const event = createMockAPIGatewayProxyEvent({}, "");
+    const mapOfCookies = getCookieValues(event);
+    expect(mapOfCookies).length(2);
+    expect(mapOfCookies!.get("test_cookie_one")).toBe("abc123");
+    expect(mapOfCookies!.get("test_cookie_two")).toBe("xyz246");
+  });
+
+  it("should return the specific value for the provided cookie name", async () => {
+    const event = createMockAPIGatewayProxyEvent({}, "");
+    const mapOfCookies = getCookieValues(event);
+    expect(mapOfCookies).length(2);
+    expect(mapOfCookies!.get("test_cookie_one")).toBe("abc123");
+  });
+
+  it("should return undefined when an error is thrown", async () => {
+    const decodeSpy = vitest.spyOn(globalThis, "decodeURIComponent").mockImplementation(() => {
+      throw new URIError("URI malformed");
+    });
+    const event = createMockAPIGatewayProxyEvent({}, "");
+    const mapOfCookies = getCookieValues(event);
+    expect(decodeSpy).toHaveBeenCalled();
+    expect(mapOfCookies).toBe(undefined);
+  });
+});
+
+const createMockAPIGatewayProxyEvent = (event: Partial<APIGatewayProxyEvent>, body: string): APIGatewayProxyEvent => ({
+  body: body,
+  headers: { Cookie: "test_cookie_one=abc123;test_cookie_two=xyz246" },
+  multiValueHeaders: {},
+  httpMethod: "POST",
+  isBase64Encoded: false,
+  path: "/",
+  // eslint-disable-next-line unicorn/no-null -- Required to create valid APIGatewayProxyEvent
+  pathParameters: null,
+  queryStringParameters: { redirect_uri: "https://api.example.com", state: "test-state", client_id: "test-client-id" },
+  // eslint-disable-next-line unicorn/no-null -- Required to create valid APIGatewayProxyEvent
+  multiValueQueryStringParameters: null,
+  // eslint-disable-next-line unicorn/no-null -- Required to create valid APIGatewayProxyEvent
+  stageVariables: null,
+  requestContext: {} as APIGatewayEventRequestContextWithAuthorizer<never>,
+  resource: "/",
+  ...event,
+});
--- a/src/handlers/get-callback-handler/get-callback-request.ts
+++ b/src/handlers/get-callback-handler/get-callback-request.ts
@@ -1,0 +1,20 @@
+export interface AuthorizationQueryStringParameters {
+  redirect_uri: string;
+  state: string;
+  client_id: string;
+}
+
+export function isValidQueryParameters(object: unknown): object is AuthorizationQueryStringParameters {
+  if (!object || typeof object !== "object") return false;
+  return (
+    "redirect_uri" in object &&
+    typeof object.redirect_uri === "string" &&
+    object.redirect_uri.trim().length > 0 &&
+    "state" in object &&
+    typeof object.state === "string" &&
+    object.state.trim().length > 0 &&
+    "client_id" in object &&
+    typeof object.client_id === "string" &&
+    object.client_id.trim().length > 0
+  );
+}
--- a/src/handlers/get-callback-handler/get-callback-response.ts
+++ b/src/handlers/get-callback-handler/get-callback-response.ts
@@ -1,0 +1,33 @@
+export interface AuthorizationSuccessResponse {
+  redirectionURI: string;
+  authorizationCode: { value: string };
+  state: { value: string };
+}
+
+export function isValidSuccessResponse(object: unknown): object is AuthorizationSuccessResponse {
+  if (!object || typeof object !== "object") return false;
+  if (!(
+    "redirectionURI" in object &&
+    typeof object.redirectionURI === "string" &&
+    object.redirectionURI.trim().length > 0
+  )) {
+    return false;
+  }
+
+  if (!("authorizationCode" in object && object.authorizationCode && typeof object.authorizationCode === "object")) {
+    return false;
+  }
+
+  if (!("state" in object && object.state && typeof object.state === "object")) {
+    return false;
+  }
+
+  return (
+    "value" in object.authorizationCode &&
+    typeof object.authorizationCode.value === "string" &&
+    object.authorizationCode.value.trim().length > 0 &&
+    "value" in object.state &&
+    typeof object.state.value === "string" &&
+    object.state.value.trim().length > 0
+  );
+}
--- a/src/handlers/get-callback-handler/tests/get-callback-handler.test.ts
+++ b/src/handlers/get-callback-handler/tests/get-callback-handler.test.ts
@@ -1,0 +1,262 @@
+import { APIGatewayEventRequestContextWithAuthorizer, APIGatewayProxyEvent } from "aws-lambda";
+import { expect, it, vitest } from "vitest";
+import { handler } from "../get-callback-handler";
+import { getCookieValues } from "../../../commons/cookie-utilities";
+
+process.env.DOMAIN_NAME = "test-domain";
+process.env.OAUTH_INTERNAL_API = "https://test.com";
+
+const { mockError } = vitest.hoisted(() => {
+  return {
+    mockError: vitest.fn(),
+  };
+});
+
+vitest.mock("@aws-lambda-powertools/logger", () => {
+  return {
+    Logger: class {
+      error = mockError;
+      constructor() {}
+    },
+  };
+});
+
+it("should return a 302 status code and redirect with an auth code and state on a successful request", async () => {
+  const event = createMockAPIGatewayProxyEvent({}, "");
+  const token = getCookieValues(event)!.get("identity_reuse_service_session");
+  expect(token).toBe("abc123");
+
+  globalThis.fetch = vitest.fn().mockResolvedValue({
+    status: 200,
+    json: vitest.fn().mockResolvedValue({
+      redirectionURI: "https://api.example.com",
+      authorizationCode: { value: "test-auth-code" },
+      state: { value: "test-state" },
+    }),
+  });
+
+  const response = await handler(event);
+  expect(response).toStrictEqual({
+    statusCode: 302,
+    body: "",
+    headers: {
+      Location: "https://api.example.com/?code=test-auth-code&state=test-state",
+    },
+  });
+});
+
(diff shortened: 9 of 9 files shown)


src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,92 @@
+import { APIGatewayProxyEvent, APIGatewayProxyResult } from "aws-lambda";
+import logger from "../../commons/logger";
+import { URL } from "node:url";
+import Cookies from "js-cookie";
+
```
umaar: In the interest of having one less dependency, would you possibly consider doing this natively, as this might be good considering the general state of supply chain issues. If our usage of cookie getting/setting is simple, this could be worth doing. Not sure what you think

umaar: Also curious how this might work as I think it only reads from `document.cookie` which is [unavailable](https://github.com/js-cookie/js-cookie/blob/main/src/api.mjs#L56) here. I appreciate this has probably been tested so I might be missing something obvious!

mislam246: Good point, agreed with using built-in functionality instead of an external dependency in this case, given it will only read a cookie value. I initially used `js-cookie` to simplify the code. The tests were working when using the package, but I've replaced it now.

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -90,3 +89,18 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
     };
   }
 };
+
+export function getCookieValueByName(name: string): string | undefined {
+  const cookieArray = document.cookie.split(";");
```
umaar: My brain is probably being very slow! I know you've tested this all but it feels a bit "magic" to me that this works, considering we're in a lambda. I personally wouldn't have done it this way, but also respect that if it works, that should be considered. Will reply on Slack

src/handlers/get-callback/get-callback-handler.ts
chrisclayson: Yeah, I don't think anything that uses `document.cookie` is going to work, as @umaar says, `document` isn't available (I don't think).

The correct way to achieve this, in Lambda, would be to parse the `event` object which should contain the headers of the request. Sadly, I don't think contains pre-parsed cookies, so you'll have to write code to parse the `Cookie` header.

umaar: Wonder if, in the future rather than now, we could consider making sure TypeScript correctly catches stuff like this. So `document` cannot be used in a lambda - I thought TypeScript would've caught that anyway so wonder if something is misconfigured. There are even linting [rules](https://eslint.org/docs/latest/rules/no-restricted-globals) which can help, so accessing assumed globals like `document` within Lambda handler code would fail.

mislam246: That's a good idea, because none of the tests were picking up this issue locally or in dev.

I've updated the code to read the cookie value from the event's header now.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: It was my bad, I misunderstood the original comment thinking it was just to replace the external dependency. I've updated the logic now to use the header instead of using `document`.

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+  if (!cookieHeader) {
+    return undefined;
+  }
+
+  const cookies = cookieHeader.split(";");
+
+  for (const element of cookies) {
+    const cookie = element.trim();
+
+    if (cookie.startsWith(name + "=")) {
```
chrisclayson: Not sure how they compare performance wise, I'd have been tempted to use `split()` here, something like:

```
const cookie = element.trim().split("=");
if (cookie[0] === name) {
  return decodeURIComponent(cookie[1]);
}
```

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+      statusCode: 302,
+      headers: {
+        Location: `https://${domainName}/error/unrecoverable`,
+      },
+      body: "",
+    };
+  }
+};
+
+export function getCookieValueByName(event: Partial<APIGatewayProxyEvent>, name: string): string | undefined {
```
chrisclayson: This might be useful to pull into shared helper library, as it's going to be required by other endpoints.

A useful refactor to this (though not necessary now) would be to, maybe, just get it to return a Map of cookies then, if the caller needs to get multiple cookies, we're not repeatedly parsing the header.

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: All good. Just to let you know even if you had kept the external dep, it still would be the same problem because that dep read from the `document`, which is unavailable in a lambda.

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
         - Key: Name
           Value: !Sub "/aws/lambda/${UnrecoverableErrorLambda}"
 
+  OauthCallbackLambda:
+    Type: AWS::Serverless::Function
+    Condition: IsNotProduction
+    Properties:
+      FunctionName: !Sub "${AWS::StackName}-OauthCallbackLambda"
+      Handler: oauth-callback-handler.handler
+      Role: !GetAtt ConfirmDetailsRole.Arn
```
umaar: Just double checking if `ConfirmDetailsRole` is correct

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+  authorizationCode: { value: string };
+  state: { value: string } | object;
+}
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const {
+    redirect_uri: redirectUri,
+    state,
+    client_id,
+  } = event.queryStringParameters as AuthorizationQueryStringParameters;
```
umaar: Wondering if possible that these are unavailable at runtime, e.g. if no query params are provided. If so, that case might need to be handled with a graceful redirect? E.g. consider that they're used in `url.searchParams.append`

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+      }
+      return {
+        statusCode: 302,
+        headers: {
+          Location: `${orchestrationRedirectUrl}`,
+        },
+        body: "",
+      };
+    } else if (responseFromAuthorizeEndpoint.status === 403) {
+      const orchestrationRedirectUrl = new URL(redirectUri);
```
umaar: Is this potentially an anti pattern because of [open redirect attack vectrors](https://owasp.org/www-community/attacks/open_redirect)? Wonder if the redirect URL needs some sort of validating

src/handlers/oauth-callback/tests/oauth-callback-handler.test.ts
```diff
@@ -0,0 +1,130 @@
+import { vitest } from "vitest";
+import { APIGatewayEventRequestContextWithAuthorizer, APIGatewayProxyEvent } from "aws-lambda";
+import { expect, it } from "vitest";
+import { handler } from "../oauth-callback-handler";
+import { getCookieValueByName } from "../oauth-callback-handler";
+
+vitest.mock("js-cookie");
```
umaar: I think this js-cookie was being removed?

openAPI/public-api.yaml
```diff
@@ -160,6 +160,18 @@ paths:
         requestParameters:
           integration.request.path.item: method.request.path.item
         type: aws
+  /oauth/callback:
```
michaelabenyohai: I think we might want to make this `/oauth2/callback` rather than `/oauth`. In other places `oauth2` seems to be used, and this might have just been a typo in the story/diagrams (probably my fault).

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+      orchestrationRedirectUrl.searchParams.append("state", state);
+      return {
+        statusCode: 302,
+        headers: {
+          Location: `${orchestrationRedirectUrl}`,
+        },
+        body: "",
+      };
+    } else {
+      return {
```
umaar: Was curious your thoughts if this `else` branch would benefit from logging

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Based on my understanding from reading the docs, the original approach using `startsWith()` and `slice()` will be faster and more memory-efficient, given it'll only create and populate an array if it matches the the cookie name passed in. With the suggested `split()` approach, the array will be created straight away and populated with all the string objects regardless.

There's also a bug with using `.split()` on cookies with `=`, where the values get shortened and it's advised to use `.slice()` instead (https://stackoverflow.com/a/10884619).

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  const cookies = cookieHeader.split(";");
+
+  for (const element of cookies) {
+    const cookie = element.trim();
+
+    if (cookie.startsWith(name + "=")) {
+      const cookieValue = cookie.slice(name.length + 1);
+      return decodeURIComponent(cookieValue);
```
umaar: Worth being mindful that this can [throw an error](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent#exceptions), and may be considered good practice to use `try/catch` just in case

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+      headers: {
+        Location: `https://${domainName}/error/unrecoverable`,
+      },
+      body: "",
+    };
+  }
+};
+
+export function getCookieValueByName(event: Partial<APIGatewayProxyEvent>, name: string): string | undefined {
+  const cookieHeader = event.headers?.["cookie"] || event.headers?.["Cookie"];
```
umaar: Does all upper case need handling I wonder? Could consider lowercasing header keys to make it easier

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,112 @@
...
+export type AuthorizationQueryStringParameters = {
+  redirect_uri: string;
+  state: string;
+  client_id: string;
+};
+
+interface AuthorizationSuccessResponse {
+  redirectionURI: string;
+  authorizationCode: { value: string };
+  state: { value: string } | object;
```
umaar: nit: Wonder if `| object` is too loose?  Feel free to ignore

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Would that go in the `commons` folder?

src/handlers/get-callback/get-callback-handler.ts
mislam246: Wouldn't that be covered by the catch block, which would redirect to the error page? The only other redirect would be back to orchestration, but if I'm not mistaken, that's only when we get a `403 access_denied` response from calling the `/api/authorization` endpoint.

src/handlers/get-callback/get-callback-handler.ts
chrisclayson: We absolutely should _but_ we don't currently have a thing to validate it against (i.e. we haven't yet built our "client registry" which will contain the configuration of our allowed clients, their ID, redirect URIs, location of JWKS endpoint etc). We'll be doing this is the next sprint.

src/handlers/get-callback/get-callback-handler.ts
chrisclayson: Agreed ... we should not be shy about logging, it will future us debug stuff 🤓

src/handlers/oauth-callback/oauth-callback-handler.ts
chrisclayson: Yes, I think so, call it something like `cookie-helper` or `cookie-utilities` 🤷🏻

src/handlers/oauth-callback/oauth-callback-handler.ts
chrisclayson: I've got a vague recollection that header names should be lower cased before encoding anyway so, technically, any upper case value is not compliant ... I might be wrong.

I've just got a memory of writing case-insensitive code (along these lines) and subsequently finding out it wasn't necessary as, by the time it reaches the lambda, all the keys are lower case anyway 🤔 

Please double check this, though, as I'm may be mis-remembering.

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+      Environment:
+        Variables:
+          NODE_OPTIONS: "--enable-source-maps"
+          DOMAIN_NAME: !Ref ReuseIdentityCustomDomain
+          OAUTH_INTERNAL_API:
+            Fn::Sub:
+              - "${OauthInternalStackId}"
+              - OauthInternalStackId:
+                  Fn::ImportValue: !Sub "${OauthInternalStackName}-OAuthInternalApi"
```
michaelabenyohai: I think this might be renamed to `PrivateApi` depending on another PR review

You won't be able to import this value in any environment other than dev until that stack is deployed all the way up, so merging this will be blocked until SPT-1922 and SPT-1950 are done.

src/handlers/oauth-callback/oauth-callback-handler.ts
chrisclayson: I'm slightly wrong, in HTTP/2 they must be lowercase, HTTP 1.x they can be either (but must be compared case insensitively) ... but I'm still convinced that they lowercased before reaching the lambda, maybe API Gateway does that!? 🤔

src/handlers/get-callback/get-callback-handler.ts
umaar: The existing try/catch would catch a fetch failure yep, but I mean before that, what if the query string params are `null` for example, which could, in theory, happen for a number of reasons.

There's also a best practice that suggests early exits, e.g. guard clauses _before_ network requests is a good idea. One thinking is: Why make a network request which you know will fail.

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: Few thoughts in no particular order:

1. I find stuff like `.startsWith` to be very readable and useful compared to older-style counterparts.
2. The original PR code is correct, whereas the suggested code does have a bug: when cookie values are base64 encoded and/or use `=` for padding, the current suggested code would fail to the read the cookie correctly.
3. Was curious about the performance suggestion, and while this is completely negligible (wouldn't be noticeable to anyone in the slightest), the original version is indeed faster by quite a bit, using those online JS benchmarking tools:

Image

src/handlers/get-callback/get-callback-handler.ts
umaar: Thanks for explaining

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: Interesting! I don't know nuances behind what API Gateway does to headers (have read a few conflicting bits online) but I would say if it doesn't cost us a lot, maybe there's no harm in lower-casing regardless. Just to be resilient, and also to not couple ourselves to what API Gateway happens to do.

src/handlers/get-callback/get-callback-handler.ts
michaelabenyohai: I agree that it would be good to have a Type Guard here. I think this is something we have missing in a few places where we parse/deserialise requests and might want to retrospectively check those. 

Agree we should do it here. The only reason not to might be that these query params are going to disappear at some point (in favour of some kind of frontend session maybe) but it's a small change and probably good to get into the habit of using type guards more.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Makes sense and thanks for the reference. Wouldn't the main `try/catch` block in the `handler` function cover any errors thrown from within this `getCookieValueByName` function?

src/handlers/get-callback/get-callback-handler.ts
michaelabenyohai: @mislam246 the thing to be aware of with TypeScript (you might know this already) is that using `as` does not throw an error if the runtime data does not match the type. In something like Java, this type of thing would throw at this point. Instead the JavaScript will just throw a runtime error further down the stack when it comes to be used, as Umar says, which isn't great.

src/handlers/get-callback/get-callback-handler.ts
mislam246: That could be of use for debugging; I've added a generic log here with the status code returned. I'm not sure if the response contains the error reason.

src/handlers/oauth-callback/tests/oauth-callback-handler.test.ts
mislam246: Missed that, removed it now.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Changed this to `state: { value?: string };`

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Changed this to convert headers to lowercase before processing, to cover all casing scenarios of the header name.

openAPI/public-api.yaml
mislam246: Renamed the endpoint to `/oauth2/callback`.

src/handlers/get-callback-handler/get-callback-handler.ts
michaelabenyohai, line 1: Can I suggest that we call this Lambda `get-callback-handler` so that it is inline with the other name changes in https://govukverify.atlassian.net/browse/SPT-1969?

Can you also make sure the directory name matches?

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
         - Key: Name
           Value: !Sub "/aws/lambda/${UnrecoverableErrorLambda}"
 
+  OauthCallbackLambda:
```
michaelabenyohai: As per another comment I suggest this and other resources should be named `GetCallbackHandler` to be consistent with the other name changes

src/handlers/get-callback/get-callback-handler.ts
mislam246: @chrisclayson / @michaelabenyohai is there a fixed list of domains we can validate against? I'm assuming if we check against those, that would be sufficient protection?

template.yaml
mislam246: Fixed it, good spot.

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: Totally ignore this, just an FYI, this could have been one alternative to cookie parsing. But nothing to action from this code snippet below, is more of an FYI to share another approach.

```js
const cookieHeader = "sessionId=abc123; theme=dark; userId=42";
const name = "theme";

const normalised = cookieHeader.replaceAll("; ", "&");
const params = new URLSearchParams(normalised);
const value = params.get(name);

console.log(value); // "dark"
```

No need to action this however!

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: Currently, I don't think it's in the handler try/catch, so wouldn't be caught in there.

Even if it were, as a general note: my suggestion is: higher-level try catches are good for unpredictable stuff, but additionally, it's good to handle errors _where they're understood_. IMO, in this example, both would be good.

Elsewhere in our products, we do have 'broad' `try`'s which swallow everything, and it can be a bit challenging during debugging.

src/handlers/get-callback/get-callback-handler.ts
umaar: I think Chris is suggesting that will be done next sprint, not too sure about where the registry comes from. I'm assuming the one thing which allows security work like this to be postponed is the `Condition: IsNotProduction` flag.

michaelabenyohai: This is one of the things that I mentioned in Slack that I think is a problem with OAuth Common.

In my opinion the `/api/authorization` endpoint should do the checking, as it is the OAuth common stack that has the concept of the client registry.

If you look at the bit of the `if` where we are responding to a `200` response from `/api/authorization`, you will nptice that it returns `redirectionURI`. That HAS been checked against the expected list of redirect URIs for the client.

Annoyingly, with the `403` response, `/api/authorization` does not return the validated `redirectionURI` which is why in this PR we are temporarily having to use the "unvalidated one".

I think `/api/authorization` SHOULD return the validated `redirectUri` on `403` response and I'm going to raise it with the CRI teams as bit of improvement we might like to make. I think this all aligns with the OAuth spec.

template.yaml
mislam246: Renamed all the resources as suggested.

src/handlers/get-callback-handler/get-callback-handler.ts
mislam246: Renamed the Lambda and the directory for consistency with the other SIS functions.

template.yaml
mislam246: Thanks for pointing it out. Updated this to append `PrivateApi` now. I'll avoid merging until those two tickets are merged and deployed successfully.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: I've refactored it slightly so the `decodeURIComponent()` is surrounded in a try/catch block and the handler would redirect to the error page.

src/handlers/get-callback/get-callback-handler.ts
mislam246: Ah, those comments weren't showing until recently for some reason. I'll leave this to be actioned in the next sprint then.

src/handlers/get-callback/get-callback-handler.ts
mislam246: Thanks both for the explanations. I've added a Type Guard to check the data types and presence of the fields, redirecting to the error page if there're any mismatches.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: That makes sense, I've moved the logic to a new `cookie-utilities` file so it can be reused by other endpoints if needed. I've also refactored it to take in a list of cookie names and return a Map of values as suggested.

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,117 @@
...
+  try {
+    const responseFromAuthorizeEndpoint = await fetch(url, {
+      method: "GET",
+      headers: {
+        "session-id": sessionId,
+      },
+    });
+
+    if (responseFromAuthorizeEndpoint.status === 200) {
+      const authorizationData = (await responseFromAuthorizeEndpoint.json()) as AuthorizationSuccessResponse;
```
michaelabenyohai: Can you extract the code here into a separate function called something like `redirectToClient`. That can then be reused in the  `else if` and remove some duplication. It would also make the if statement much more readable if it was:

```
if (responseFromAuthorizeEndpoint.status === 200) {
  redirectToClient(...)
} else if (responseFromAuthorizeEndpoint.status === 403) {
  redirectToClient(...)
} else {
  redirectToErrorPage(...)
}
```

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Thanks for the references. Based on your comments @umaar, I assume its preferred to leave this as is?

src/handlers/oauth-callback/oauth-callback-handler.ts
```diff
@@ -0,0 +1,117 @@
...
+      }
+      return {
+        statusCode: 302,
+        headers: {
+          Location: `${orchestrationRedirectUrl}`,
+        },
+        body: "",
+      };
+    } else if (responseFromAuthorizeEndpoint.status === 403) {
+      const orchestrationRedirectUrl = new URL(redirectUri);
```
michaelabenyohai: As above, could you extract this stuff out into a function called `redirectToErrorPage()`. It would make this `if` statement much easier to understand when skimming over it at a later date.

src/handlers/get-callback/get-callback-handler.ts
chrisclayson: Yeah, it also occurred to me, last night, that the `redirectUri` query param on the callback is actually just a fudge, for the moment, to keep the stub journey working.

As @michaelabenyohai says this would actually be retrieved from the session in the 'real' journey.

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+
+    if (responseFromAuthorizeEndpoint.status === 200) {
+      const authorizationData = (await responseFromAuthorizeEndpoint.json()) as AuthorizationSuccessResponse;
+      const orchestrationRedirectUrl = new URL(decodeURIComponent(authorizationData.redirectionURI));
+      orchestrationRedirectUrl.searchParams.append("code", authorizationData.authorizationCode.value);
+      if (authorizationData.state?.value) {
+        orchestrationRedirectUrl.searchParams.append("state", authorizationData.state.value);
+      } else {
+        orchestrationRedirectUrl.searchParams.append("state", "");
```
michaelabenyohai: If there is no `state` value we must not redirect to Orch as the `state` is required. 

In this case can you redirect to the error page?

From RFC0014:
> The following situations do not result in an OAuth error response; instead the user would just see an error message in an HTTP 500 response:
> * problem with the signature on the JAR;
> * problem decrypting the JAR;
> * invalid redirect_uri;
> * any other error that prevents IPV Core constructing a valid return redirect including the original state value.

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+export type AuthorizationQueryStringParameters = {
+  redirect_uri: string;
+  state: string;
+  client_id: string;
+};
+
+interface AuthorizationSuccessResponse {
+  redirectionURI: string;
+  authorizationCode: { value: string };
+  state: { value?: string };
```
michaelabenyohai: Should `value` be required here? I don't think we would ever want it to be optional.

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+  try {
+    const responseFromAuthorizeEndpoint = await fetch(url, {
+      method: "GET",
+      headers: {
+        "session-id": sessionId,
+      },
+    });
+
+    if (responseFromAuthorizeEndpoint.status === 200) {
+      const authorizationData = (await responseFromAuthorizeEndpoint.json()) as AuthorizationSuccessResponse;
```
michaelabenyohai: Please can you implement a Type Guard here? 

If the structure of the `authorizationData` does not match the `AuthorizationSuccessResponse` I think we should go to the error page.

src/handlers/get-callback/get-callback-handler.ts
michaelabenyohai: This could be handled by the Type Guard I suggest implementing in the comment above

template.yaml
```diff
@@ -81,6 +81,11 @@ Parameters:
     Type: String
     Default: vpc
 
+  OauthInternalStackName:
+    Description: OAuth Internal Stack name
+    Type: String
+    Default: "oauth-internal"
```
michaelabenyohai: This will vary by environment. In build-> prod we have agreed that this will be `sis-oauth-internal`.

In dev/local I think this will be `preview-oauth-internal`.

@chrisclayson is it worth putting these values in the `EnvironmentMap`?  Or default the value here to something (maybe `sis-oauth-internal` and override it for dev in the GitHub Actions?

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
         - Key: Name
           Value: !Sub "/aws/lambda/${UnrecoverableErrorLambda}"
 
+  GetCallbackLambda:
```
michaelabenyohai: Can you replace `GetCallbackLambda` with  `GetCallbackHandler` everywhere?

Sorry, there is a PR renaming everything to use that pattern so would be good to stick to it.

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+        EntryPoints:
+          - src/handlers/get-callback/get-callback-handler.ts
+        Format: esm
+        Platform: node
+        OutExtension:
+          - .js=.mjs
+        Banner:
+          - js=import { createRequire } from 'module'; const require = createRequire(import.meta.url);
+
+  GetCallbackRole:
```
michaelabenyohai: Can this be `GetCallbackHandlerRole` to align with new names too

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - "sts:AssumeRole"
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
+
+  GetCallbackLambdaRestApiRole:
```
michaelabenyohai: Similarly `GetCallbackHandlerRestApiRole`

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+        - Key: System
+          Value: !Ref SystemTagValue
+        - Key: Environment
+          Value: !Ref Environment
+        - Key: Owner
+          Value: !Ref OwnerTagValue
+        - Key: Source
+          Value: !Ref SourceTagValue
+
+  GetCallbackLambdaRestApiPolicy:
```
michaelabenyohai: Similarly `GetCallbackHandlerRestApiPolicy`

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+        Statement:
+          - Effect: "Allow"
+            Action:
+              - "lambda:InvokeFunction"
+            Resource:
+              - !Sub "arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:${AWS::StackName}-GetCallbackLambda:*"
+      Roles:
+        - !Ref GetCallbackLambdaRestApiRole
+
+  GetCallbackFunctionLogGroup:
```
michaelabenyohai: Similarly `GetCallbackHandlerLogGroup`

template.yaml
```diff
@@ -1420,6 +1425,129 @@ Resources:
...
+        - Key: System
+          Value: !Ref SystemTagValue
+        - Key: Environment
+          Value: !Ref Environment
+        - Key: Owner
+          Value: !Ref OwnerTagValue
+        - Key: Source
+          Value: !Ref SourceTagValue
+        - Key: Name
+          Value: !Sub "/aws/lambda/${GetCallbackLambda}"
```
michaelabenyohai: Similarly `GetCallbackHandler`

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,106 @@
...
+  redirectionURI: string;
+  authorizationCode: { value: string };
+  state: { value?: string };
+}
+
+export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayProxyResult> => {
+  const {
+    redirect_uri: redirectUri,
+    state,
+    client_id,
```
michaelabenyohai: We seem to use camelCase for `redirectUri` and snake_case for `client_id`. I wonder if we should be consistent, and probably use camelCase for both?

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Refactored this logic into separate functions as suggested. I've also separated out the client redirect ones with/without an error, otherwise we'd have to pass in the status code to the function.

src/handlers/oauth-callback/oauth-callback-handler.ts
mislam246: Done as above.

src/handlers/get-callback/get-callback-handler.ts
mislam246: The new `isValidQueryParameters` Type Guard function should cover this now.

src/handlers/get-callback/get-callback-handler.ts
mislam246: I've updated this to make the value mandatory.

src/handlers/get-callback/get-callback-handler.ts
mislam246: I've added a Type Guard for the response now. Do you think the new function is a bit convoluted? I included the different checks on the data types and presence of all the properties.

template.yaml
mislam246: ~~I've added these values to the EnvironmentMap for now and used them to set the default value of this parameter.~~

**Edit**: Reading `EnvironmentMap` values from the Parameters doesn't work. From what I've read, it's not possible to use Mappings or an If statement to set the Default value of Parameters.

src/handlers/get-callback/get-callback-handler.ts
mislam246: I've removed this object and am getting those values directly from `event.queryStringParameters`. The Type Guard should validate those values now.

template.yaml
mislam246: Sorry, I was sloppy with the renaming above and missed the endings. I've updated this and all other instances to be consistent with the SIS function renaming change.

template.yaml
mislam246: Changed, as above.

template.yaml
mislam246: Changed, as above.

template.yaml
mislam246: Changed, as above.

template.yaml
mislam246: Changed, as above.

template.yaml
mislam246: Changed, as above.

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: That would be my suggestion yeah, but want to respect that @chrisclayson made the original suggestion so maybe he can share his thoughts

src/handlers/get-callback/get-callback-handler.ts
umaar: Thank you

src/handlers/oauth-callback/oauth-callback-handler.ts
umaar: Thank you

template.yaml
michaelabenyohai: Can you default this to `sis-oauth-internal`?

Then you need to add the parameter override for dev in the GitHub action - add `preview-main-oauth` to this line https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/af5354230ee56199978ed82fb756d61dfefca7fd/.github/workflows/deploy-to-aws.yml#L132

Note that we had to change it to `preview-main-oauth` as `preview-main-oauth-internal` was too long for AWS limits.

mislam246: I've updated the default value. Also overrode it in the `deploy-to-aws` GH action and `deploy-to-dev.sh` script.

src/handlers/get-callback/get-callback-handler.ts
michaelabenyohai: It's quite long but I think it has to be.

I think the best thing to do is to move the types and associated type guards into their own files, then import them into the handler to make use of them. This will make the handler easier to understand as it will be specific to the logic of redirecting rather than including all the type code.

As an example, we have out types split into their own files in this directory: https://github.com/govuk-one-login/ipv-identity-reuse-service/tree/main/src/handlers/user-identity

Screenshot 2026-07-06 at 11 58 59

The handler and the types (including type guards) are separated out

template.yaml
michaelabenyohai: No worries!

src/handlers/get-callback-handler/get-callback-handler.ts
michaelabenyohai: It looks like the directory is still called `get-callback`, can you rename it to `get-callback-handler`, this will be consistent once https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/363 is merged

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,155 @@
...
+      return await redirectToClient(responseFromAuthorizeEndpoint, domainName);
+    } else if (responseFromAuthorizeEndpoint.status === 403) {
+      return await redirectToClientWithError(queryParameters.redirect_uri, queryParameters.state);
+    } else {
+      logger.error(`${responseFromAuthorizeEndpoint.status} response code returned from the authorization endpoint`);
+      return await redirectToErrorPage(domainName);
+    }
+  } catch (error) {
+    logger.error(`Error in OAuth Callback handler event: ${error}`);
+    return {
```
michaelabenyohai: This can be replaced with `return await redirectToErrorPage(domainName);`

src/handlers/get-callback/get-callback-handler.ts
```diff
@@ -0,0 +1,155 @@
...
+  const orchestrationRedirectUrl = new URL(decodeURIComponent(authorizationData.redirectionURI));
+  orchestrationRedirectUrl.searchParams.append("code", authorizationData.authorizationCode.value);
+  orchestrationRedirectUrl.searchParams.append("state", authorizationData.state.value);
+  return {
+    statusCode: 302,
+    headers: {
+      Location: `${orchestrationRedirectUrl}`,
+    },
+    body: "",
+  };
```
michaelabenyohai: Lines 78 to 84 are duplicated 3 times in this file. Can you extract a function called something like: `redirect(location: string)`

src/handlers/get-callback/get-callback-handler.ts
mislam246: Missed this one. Changed it to call the new function.

src/handlers/get-callback/get-callback-handler.ts
mislam246: Refactored into a new function and removed the duplication.

src/handlers/get-callback/get-callback-handler.ts
mislam246: Makes sense. I've moved the types into separate files along with the type guards.

deploy-to-dev.sh
```diff
@@ -69,7 +69,7 @@ deploy_or_destroy() {
         --no-fail-on-empty-changeset \
         --profile "$AWS_PROFILE" \
         --tags DeploymentSource=Manual StackType=Dev Project="ipv-identity-reuse-service" \
-        --parameter-overrides Environment="$ENVIRONMENT" $PARAMETER_OVERRIDES
+        --parameter-overrides Environment="$ENVIRONMENT" OauthInternalStackName="preview-oauth-internal" $PARAMETER_OVERRIDES
```
michaelabenyohai: You don't need this change and it can be removed. This already exists on line 192 (a previous PR introduced it)

michaelabenyohai (dismissed):

deploy-to-dev.sh
mislam246: Removed this change.

src/handlers/oauth-callback/oauth-callback-handler.ts
chrisclayson: Interesting, I think the `=` and base64 'bug' is a consideration. Personally, I think we shouldn't be directly setting cookie values with any value not URL safe in there, I believe it is common practice to URL encode cookies BUT this isn't actually required by the RFC, so we should make it safe as we can't be sure what a malicious actor could do with this bug.

My main concern was about having something re-usable that would be equally performant if a single function needs to read multiple cookies. I was trying to avoid multiple iterations of the header, by parsing once into a map (which is fairly common in frameworks I've used). That said we're only ever likely to have 2, maybe 3, cookies so it may not be a massive consideration.

mislam246: I've refactored the logic by removing the `.split()` from within the for loop and replaced it with a `slice()`, using the index of the `=` character.

Also modified the function behaviour to return the entire Map of cookies and the handler now uses `Map.get()` to retrieve a cookie value by name.

chrisclayson (approved):

michaelabenyohai (approved):

PR #367, opened 2026-06-29
SPT-1937: Validate JAR on `/authorize`

umaar:
## What

- Validate JAR and create session on call to /authorize

## Why

- In order to enable SIS to make use of Common OAuth components, SIS and it’s Orchestration stub will require updates to integrate into the OAuth flow.

### DIFF
--- a/.github/workflows/deploy-to-aws.yml
+++ b/.github/workflows/deploy-to-aws.yml
@@ -124,19 +124,19 @@ jobs:
           stack-names: ${{ env.STACK_NAME }}
           aws-region: ${{ env.AWS_REGION }}
 
-#      N.B. Comment out until we've solved the circular dependency
-#      - name: Deploy OAuth Stack
-#        timeout-minutes: 15
-#        run: |
-#          echo "$OAUTH_STACK_NAME"
-#          sam deploy \
-#          --stack-name "$OAUTH_STACK_NAME"
-#          --region "$AWS_REGION"  \
-#          --no-confirm-changeset \
-#          --no-fail-on-empty-changeset \
-#          --capabilities CAPABILITY_NAMED_IAM,CAPABILITY_AUTO_EXPAND \
-#          --tags DeploymentSource="GitHub Actions" StackType=Preview \
-#          --parameter-overrides Environment="dev" SisStackName=preview-main
+      #      N.B. Comment out until we've solved the circular dependency
+      #      - name: Deploy OAuth Stack
+      #        timeout-minutes: 15
+      #        run: |
+      #          echo "$OAUTH_STACK_NAME"
+      #          sam deploy \
+      #          --stack-name "$OAUTH_STACK_NAME"
+      #          --region "$AWS_REGION"  \
+      #          --no-confirm-changeset \
+      #          --no-fail-on-empty-changeset \
+      #          --capabilities CAPABILITY_NAMED_IAM,CAPABILITY_AUTO_EXPAND \
+      #          --tags DeploymentSource="GitHub Actions" StackType=Preview \
+      #          --parameter-overrides Environment="dev" SisStackName=preview-main
 
       - name: Deploy SAM Application
         timeout-minutes: 15
@@ -154,7 +154,7 @@ jobs:
             --no-fail-on-empty-changeset \
             --capabilities CAPABILITY_NAMED_IAM \
             --tags DeploymentSource="GitHub Actions" StackType=Preview \
-            --parameter-overrides Environment="dev"
+            --parameter-overrides Environment="dev" VpcStackName=vpc OauthInternalStackName=preview-main-oauth
 
       - name: Report deployment
         run: echo "Deployed stack \`$STACK_NAME\`" >> "$GITHUB_STEP_SUMMARY"
--- a/.github/workflows/deploy-to-aws.yml
+++ b/.github/workflows/deploy-to-aws.yml
@@ -154,7 +154,7 @@ jobs:
             --no-fail-on-empty-changeset \
             --capabilities CAPABILITY_NAMED_IAM \
             --tags DeploymentSource="GitHub Actions" StackType=Preview \
-            --parameter-overrides Environment="dev" VpcStackName=vpc OauthInternalStackName=preview-main-oauth
+            --parameter-overrides Environment="dev" OauthInternalStackName=preview-main-oauth
 
       - name: Report deployment
         run: echo "Deployed stack \`$STACK_NAME\`" >> "$GITHUB_STEP_SUMMARY"
--- a/deploy-to-dev.sh
+++ b/deploy-to-dev.sh
@@ -165,7 +165,7 @@ if $DEPLOY_SIS; then
 
     if $DEPLOY_OAUTH; then
         OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
-    else
+    elif
         OAUTH_STACK_NAME="preview-main-oauth"
     fi
 elif $DEPLOY_OAUTH; then
--- a/deploy-to-dev.sh
+++ b/deploy-to-dev.sh
@@ -162,7 +162,12 @@ fi
 
 if $DEPLOY_SIS; then
   SIS_STACK_NAME=$STACK_NAME
-  OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
+
+    if $DEPLOY_OAUTH; then
+        OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
+    else
+        OAUTH_STACK_NAME="preview-main-oauth"
+    fi
 elif $DEPLOY_OAUTH; then
   OAUTH_STACK_NAME=$STACK_NAME
 else
@@ -184,7 +189,7 @@ $DEPLOY_OAUTH && echo "Deploy oauth-internal to stack $OAUTH_STACK_NAME, pointin
 echo
 
 if $DEPLOY_SIS; then
-  deploy_or_destroy "identity-reuse-service" "$SIS_STACK_NAME"
+  deploy_or_destroy "identity-reuse-service" "$SIS_STACK_NAME" "OauthInternalStackName=$OAUTH_STACK_NAME"
 fi
 if $DEPLOY_OAUTH; then
   deploy_or_destroy "oauth-internal" "$OAUTH_STACK_NAME" "SisStackName=$SIS_STACK_NAME"
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -81,6 +81,11 @@ Parameters:
     Type: String
     Default: vpc
 
+  OauthInternalStackName:
+    Description: OAuth Internal Stack name
+    Type: String
+    Default: "oauth-internal"
+
   DynatraceNodeLayerArn:
     Type: String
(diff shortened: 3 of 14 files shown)


src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -16,12 +11,61 @@ export type AuthorizationQueryStringParameters = {
...
 
+type SessionResponse = {
+  session_id: string;
+  state: string;
+  redirect_uri: string;
+};
+
+const SESSION_TIMEOUT_MS = 10_000;
+
+const getRequiredEnvironment = (name: string): string => {
```
michaelabenyohai: It feels to me that this function could live somewhere in the `commons` directory so that it can be reused in other handlers?

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -16,12 +11,61 @@ export type AuthorizationQueryStringParameters = {
...
+  const value = process.env[name];
+  if (!value) {
+    throw new Error(`Environment variable ${name} must be defined`);
+  }
+  return value;
+};
+
+const buildSessionCookie = (sessionId: string): string =>
```
michaelabenyohai: I wonder if this also lives in `commons` in some kind of `cookies.ts` file. I know @mislam246 is doing the other side of this (reading from the cookie) which also lives in `commons`. Maybe you could coordinate do that you don't get conflicts?

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -16,12 +11,61 @@ export type AuthorizationQueryStringParameters = {
...
+  `identity_reuse_service_session=${sessionId}; Path=/; Secure; HttpOnly; SameSite=Lax`;
+
+const redirectToError = (domainName: string): APIGatewayProxyResult => {
```
michaelabenyohai: I think this could live in some new `services/error-service.ts` (or some much better name). But basically I think this could be extracted from this handler.

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -16,12 +11,61 @@ export type AuthorizationQueryStringParameters = {
...
+  return {
+    statusCode: 302,
+    body: "",
+    headers: {
+      Location: url.href,
+    },
+  };
+};
+
+const callSessionApi = async (sessionApiUrl: string, clientId: string, request: string): Promise<Response> => {
```
michaelabenyohai: I also think this could live in `services/ouath-internal-service.ts` or something.

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -35,3 +79,37 @@ export const handler = async (event: APIGatewayProxyEvent, context: Context): Pr
     },
   };
 };
+
+const handleSessionCreation = async (
```
michaelabenyohai: I Wonder if this might be simpler as just `createSession`? I don't mind though.

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -379,6 +384,8 @@ Resources:
           SQS_AUDIT_EVENT_QUEUE_URL: !Ref AuditEventQueue
           COMPONENT_ID: !Ref ComponentId
           DOMAIN_NAME: !Ref ReuseIdentityCustomDomain
+          SESSION_API_URL:
+            Fn::ImportValue: !Sub "${OauthInternalStackName}-OAuthInternalApi"
```
michaelabenyohai: Just an FYI, this going to become `PrivateApi` rather than `OAuthInternalApi` based on another PR review.

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -81,6 +81,11 @@ Parameters:
     Type: String
     Default: vpc
 
+  OauthInternalStackName:
+    Description: OAuth Internal Stack name
+    Type: String
+    Default: "oauth-internal"
```
michaelabenyohai: See https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/364/changes#r3499371604

src/commons/environment.d.ts
```diff
@@ -3,10 +3,12 @@ declare namespace NodeJS {
   interface ProcessEnv {
     APP_CONFIG_APPLICATION: string;
     APP_CONFIG_NAME: string;
+    DOMAIN_NAME: string;
     ENVIRONMENT: string;
     EVCS_API_KEY_SECRET_ARN: string;
     POWERTOOLS_METRICS_NAMESPACE: string;
     POWERTOOLS_SERVICE_NAME: string;
+    SESSION_API_URL: string;
```
michaelabenyohai: We've called this the "oauth internal API" in other places. I wonder if this should be `OAUTH_INTERNAL_API_URL`. Or maybe we could chat as a team about whether "session API" makes more sense. Or "OAuth session API".

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -35,3 +79,37 @@ export const handler = async (event: APIGatewayProxyEvent, context: Context): Pr
...
+    logger.error(`Error calling session handler: ${error}`);
+    return redirectToError(domainName);
+  }
+
+  if (response.status !== 201) {
+    logger.error(`Session handler returned non-201 status: ${response.status}`);
+    return redirectToError(domainName);
+  }
+
+  const { session_id: sessionId }: SessionResponse = await response.json();
```
michaelabenyohai: Maybe extract this line and below to a function called `redirectToConfirmDetails()` or something?

I like the idea of being able to skim this function and seeing `redirectToError` and `redirectToConfirmDetails` and quickly get what it's doing. At the moment I have to parse several lines of code to work the second redirect out.

src/handlers/authorize-handler/authorize-handler.ts
michaelabenyohai: As per other conversations, happy for this to be done after this is merged in another PR. So will resolve for now.

src/handlers/authorize-handler/authorize-handler.ts
michaelabenyohai: As per other conversations, happy for this to be done after this is merged in another PR. So will resolve for now.

src/handlers/authorize-handler/authorize-handler.ts
michaelabenyohai: As per other conversations, happy for this to be done after this is merged in another PR. So will resolve for now.

src/handlers/authorize-handler/authorize-handler.ts
umaar: Done!

src/handlers/authorize-handler/authorize-handler.ts
umaar: Done!

src/commons/environment.d.ts
umaar: Done, `OAUTH_INTERNAL_API_URL` is good thanks

src/handlers/authorize-handler/authorize-handler.ts
umaar: Should've been like that from the start, good catch, thanks. Done

infrastructure/identity-reuse-service/template.yaml
umaar: Silly q but is there anything to action here?

infrastructure/identity-reuse-service/template.yaml
michaelabenyohai: Please can you change this to `!Sub "${OauthInternalStackName}-PrivateApi"` now? It's definitely called that, as the PR I mentioned has been merged.

infrastructure/identity-reuse-service/template.yaml
michaelabenyohai: There is now, yes.

Please can you update this `Default` value to `sis-oauth-internal` as this is what the pipeline is going to name the stack from Build -> Production.

In Dev, however, it is going to be called `preview-main-oauth` (adding "internal" made it too long). Please can you override that value in the GitHub action https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/af5354230ee56199978ed82fb756d61dfefca7fd/.github/workflows/deploy-to-aws.yml#L132

`--parameter-overrides Environment="dev" VpcStackName=vpc OauthInternalStackName=preview-main-oauth`

src/handlers/authorize-handler/authorize-handler.ts
```diff
@@ -16,13 +12,81 @@ export type AuthorizationQueryStringParameters = {
...
+    body: "",
+    headers: {
+      Location: url.href,
+      "Set-Cookie": buildSessionCookie(sessionId),
+    },
+  };
+}
+
+async function callSessionApi(apiUrl: string, clientId: string, request: string): Promise<Response> {
+  const url = new URL("/api/session", apiUrl);
```
michaelabenyohai: I did some debugging and it turns out that:

* the value of `apiUrl` is `https://some-value.execute-api.com/v1` (the value of the environment variable)
* using this `new URL(url, base)` constructor generates the `url.href` value as `https://some-value.execute-api.com/api/session`
* when it should be `https://some-value.execute-api.com/v1/api/session`

So the `/v1` is accidentally stripped out. This is what is causing the `403` response from the API call.

infrastructure/identity-reuse-service/template.yaml
umaar: Done, thanks

infrastructure/identity-reuse-service/template.yaml
umaar: Done, thank you for explaining

src/handlers/authorize-handler/authorize-handler.ts
umaar: Thank you so much for diagnosing this, and figuring all of that out. Fixed.

deploy-to-dev.sh
```diff
@@ -162,7 +162,12 @@ fi
 
 if $DEPLOY_SIS; then
   SIS_STACK_NAME=$STACK_NAME
-  OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
+
+    if $DEPLOY_OAUTH; then
+        OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
+    else
```
michaelabenyohai: I've just realised this should be `elif`

umaar: Fixed

.github/workflows/deploy-to-aws.yml
```diff
@@ -154,7 +154,7 @@ jobs:
             --no-fail-on-empty-changeset \
             --capabilities CAPABILITY_NAMED_IAM \
             --tags DeploymentSource="GitHub Actions" StackType=Preview \
-            --parameter-overrides Environment="dev"
+            --parameter-overrides Environment="dev" VpcStackName=vpc OauthInternalStackName=preview-main-oauth
```
michaelabenyohai: It looks like `VpcStackName=vpc` snuck back in during the rebase. This was removed by one of the commits on the main branch and we don't need to add it back in.

umaar: Thanks, removed

michaelabenyohai (approved):

PR #374, opened 2026-07-07
SPT-1949: Fix skeleton OAuth acceptance tests

joe-mcgarry:
## What

This adds some extra steps to the `authorization-workflow.feature` acceptance tests.
- Use of `I` have been replaced with either `the client` or `the user`
- State is now passed in along with the initial redirect URI
- User clicking the continue button is accounted for by sending a POST request (handled in `confirmDetailsSubmission()`)
- `redirectUri` and `state` have been added to the WorldDefinition to allow usage between steps

## Why

The feature tests didn't provide a robust test to give confidence in the new OAuth flow. The changes should better mimic how a real user will navigate the OAuth flow

### DIFF
--- a/tests/acceptance/features/authorization-workflow.feature
+++ b/tests/acceptance/features/authorization-workflow.feature
@@ -1,10 +1,12 @@
 Feature: authorization-workflow.feature
 
   Scenario: I should be able to complete an authorization flow and request my user-identity
-    Given I have a user profile
-    When I call the authorize endpoint, with the redirect URI "https://api.example.com"
-    Then I will be issued with an authorization code and redirected to the confirm details page
-    When I call the token endpoint with the authorization code
-    Then I will be issued with an authorization token
-    When I call the user-identity endpoint
-    Then I will be issued with my user-identity
+    Given a user has a profile
+    When the client calls the authorize endpoint, with the redirect URI "https://api.example.com" and state "sample-state"
+    Then the user will be redirected to the confirm details page
+    When the user clicks Continue
+    Then the user will be redirected to the client's redirect URI with an authorization code
+    When the client calls the token endpoint with the authorization code
+    Then the client will be issued with an authorization token
+    When the client calls the user-identity endpoint with the authorization token
+    Then the user-identity will be returned to the client
--- a/tests/acceptance/features/authorization-workflow.feature
+++ b/tests/acceptance/features/authorization-workflow.feature
@@ -5,7 +5,7 @@ Feature: authorization-workflow.feature
     When the client calls the authorize endpoint, with the redirect URI "https://api.example.com" and state "sample-state"
     Then the user will be redirected to the confirm details page
     When the user clicks Continue
-    Then the user will be redirected to the client's redirect URI with an authorization code
+    Then the user will be redirected to the client's redirect URI with an authorization code and the state
     When the client calls the token endpoint with the authorization code
     Then the client will be issued with an access token
     When the client calls the user-identity endpoint with the access token
--- a/tests/acceptance/features/authorization-workflow.feature
+++ b/tests/acceptance/features/authorization-workflow.feature
@@ -7,6 +7,6 @@ Feature: authorization-workflow.feature
     When the user clicks Continue
     Then the user will be redirected to the client's redirect URI with an authorization code
     When the client calls the token endpoint with the authorization code
-    Then the client will be issued with an authorization token
-    When the client calls the user-identity endpoint with the authorization token
+    Then the client will be issued with an access token
+    When the client calls the user-identity endpoint with the access token
     Then the user-identity will be returned to the client
--- a/tests/acceptance/steps/authorization.step.ts
+++ b/tests/acceptance/steps/authorization.step.ts
@@ -40,6 +40,7 @@ Then<WorldDefinition>("the user will be redirected to the confirm details page",
   assertRedirectResponse(this.redirectResponse);
   assert.ok(isRedirectResponse(this.redirectResponse));
   assert.equal(this.redirectResponse.origin, domainName);
+  assert.equal(this.redirectResponse.pathname, "/confirm-details");
 });
 
 When<WorldDefinition>("the user clicks Continue", async function () {
--- a/tests/acceptance/steps/authorization.step.ts
+++ b/tests/acceptance/steps/authorization.step.ts
@@ -3,6 +3,7 @@ import { WorldDefinition } from "./base-verbs.step";
 import {
   AuthorizationResponseType,
   authorize,
+  confirmDetailsSubmission,
   isAuthorizationResponse,
   isTokenResponse,
   token,
@@ -11,37 +12,49 @@ import {
 import assert from "node:assert";
 import { sisGetUserIdentityHandler } from "./utils/sis-api";
 
-Given<WorldDefinition>("I have a user profile", function () {
+Given<WorldDefinition>("a user has a profile", function () {
   // Do nothing, this is currently a placeholder with a verb to make it clear
   // what may be happening. In the future it's expected that this will set up
   // user credentials.
 });
 
 When<WorldDefinition>(
-  "I call the authorize endpoint, with the redirect URI {string}",
-  async function (redirectUri: string) {
+  "the client calls the authorize endpoint, with the redirect URI {string} and state {string}",
+  async function (redirectUri: string, state: string) {
+    this.redirectUri = redirectUri;
+    this.state = state;
     this.authorizationResponse = await authorize({
       client_id: "acceptance-tests",
-      state: "acceptance-tests",
+      state: state,
       response_type: AuthorizationResponseType.code,
       redirect_uri: redirectUri,
     });
   }
 );
 
+Then<WorldDefinition>("the user will be redirected to the confirm details page", function () {
+  const domainName = process.env.DOMAIN_NAME;
+  if ("origin" in this.authorizationResponse!) {
+    this.authorizationResponse.origin = domainName!;
+  }
+  assert.ok(isAuthorizationResponse(this.authorizationResponse));
+  assert.equal(this.authorizationResponse.origin, domainName);
+});
+
+When<WorldDefinition>("the user clicks Continue", async function () {
+  this.authorizationResponse = await confirmDetailsSubmission(this.redirectUri!, this.state!);
+});
+
 Then<WorldDefinition>(
-  "I will be issued with an authorization code and redirected to the confirm details page",
-  function () {
-    const domainName = process.env.DOMAIN_NAME;
-    if ("origin" in this.authorizationResponse!) {
-      this.authorizationResponse.origin = domainName!;
-    }
+  "the user will be redirected to the client's redirect URI with an authorization code",
+  async function () {
     assert.ok(isAuthorizationResponse(this.authorizationResponse));
-    assert.equal(this.authorizationResponse.origin, domainName);
+    assert.equal(this.authorizationResponse.origin, "https://api.example.com");
+    assert.ok(this.authorizationResponse.code, "Expected an authorization code but got none");
   }
 );
 
-When<WorldDefinition>("I call the token endpoint with the authorization code", async function () {
+When<WorldDefinition>("the client calls the token endpoint with the authorization code", async function () {
   assert.ok(isAuthorizationResponse(this.authorizationResponse));
   this.tokenResponse = await token({
     grant_type: TokenGrantType.AuthorizationCode,
@@ -49,13 +62,13 @@ When<WorldDefinition>("I call the token endpoint with the authorization code", a
   });
 });
 
-Then<WorldDefinition>("I will be issued with an authorization token", function () {
+Then<WorldDefinition>("the client will be issued with an authorization token", function () {
   assert.ok(isTokenResponse(this.tokenResponse));
   assert.ok(this.tokenResponse.access_token);
   assert.equal(this.tokenResponse.token_type, "Bearer");
 });
 
-When<WorldDefinition>("I call the user-identity endpoint", async function () {
+When<WorldDefinition>("the client calls the user-identity endpoint with the authorization token", async function () {
   assert.ok(isTokenResponse(this.tokenResponse));
 
   this.userIdentityPostResponse = await sisGetUserIdentityHandler(
@@ -67,7 +80,7 @@ When<WorldDefinition>("I call the user-identity endpoint", async function () {
   );
 });
 
-Then<WorldDefinition>("I will be issued with my user-identity", function () {
+Then<WorldDefinition>("the user-identity will be returned to the client", function () {
   assert.ok(this.userIdentityPostResponse);
   assert.deepEqual(this.userIdentityPostResponse.body, {
     sub: "urn:fdc:gov.uk:2022:TEST_USER-7B96ScRg2a-k7fN-u-sZbEjbB3hQ6gf6SM0x",
--- a/tests/acceptance/steps/authorization.step.ts
+++ b/tests/acceptance/steps/authorization.step.ts
@@ -62,13 +62,13 @@ When<WorldDefinition>("the client calls the token endpoint with the authorizatio
   });
 });
 
-Then<WorldDefinition>("the client will be issued with an authorization token", function () {
+Then<WorldDefinition>("the client will be issued with an access token", function () {
   assert.ok(isTokenResponse(this.tokenResponse));
   assert.ok(this.tokenResponse.access_token);
   assert.equal(this.tokenResponse.token_type, "Bearer");
 });
 
-When<WorldDefinition>("the client calls the user-identity endpoint with the authorization token", async function () {
+When<WorldDefinition>("the client calls the user-identity endpoint with the access token", async function () {
   assert.ok(isTokenResponse(this.tokenResponse));
 
   this.userIdentityPostResponse = await sisGetUserIdentityHandler(
--- a/tests/acceptance/steps/authorization.step.ts
+++ b/tests/acceptance/steps/authorization.step.ts
@@ -7,11 +7,13 @@ import {
   isAuthorizationResponse,
   isRedirectResponse,
   isTokenResponse,
+  RedirectResponse,
   token,
   TokenGrantType,
 } from "./utils/auth-api";
 import assert from "node:assert";
 import { sisGetUserIdentityHandler } from "./utils/sis-api";
+import { CloudFormationOutputs, getCloudFormationOutput } from "./utils/cloudformation";
 
 Given<WorldDefinition>("a user has a profile", function () {
   // Do nothing, this is currently a placeholder with a verb to make it clear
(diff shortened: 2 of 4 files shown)


tests/acceptance/features/authorization-workflow.feature
```diff
@@ -1,10 +1,12 @@
...
-    Then I will be issued with an authorization token
-    When I call the user-identity endpoint
-    Then I will be issued with my user-identity
+    Given a user has a profile
+    When the client calls the authorize endpoint, with the redirect URI "https://api.example.com" and state "sample-state"
+    Then the user will be redirected to the confirm details page
+    When the user clicks Continue
+    Then the user will be redirected to the client's redirect URI with an authorization code
+    When the client calls the token endpoint with the authorization code
+    Then the client will be issued with an authorization token
```
michaelabenyohai: Oops this may have been a typo in the story and it should be "access token" rather than "authorization token" (also needs updating in the line below)

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
       response_type: AuthorizationResponseType.code,
       redirect_uri: redirectUri,
     });
   }
 );
 
+Then<WorldDefinition>("the user will be redirected to the confirm details page", function () {
+  const domainName = process.env.DOMAIN_NAME;
+  if ("origin" in this.authorizationResponse!) {
+    this.authorizationResponse.origin = domainName!;
```
michaelabenyohai: I know this code already existed and you have just refactored it. But I am wondering why this if-statement is here at all. It appears to be that doesn't need to be, because `this.authorizationResponse.origin` should have been set by the previous step (line 26).

I feel like this if-statement means that this test would always pass. Do you think we should remove it? I can't work out why it was here in the first place.

I think the test would pass without it?

tests/acceptance/features/authorization-workflow.feature
```diff
@@ -1,10 +1,12 @@
...
-    Then I will be issued with an authorization code and redirected to the confirm details page
-    When I call the token endpoint with the authorization code
-    Then I will be issued with an authorization token
-    When I call the user-identity endpoint
-    Then I will be issued with my user-identity
+    Given a user has a profile
+    When the client calls the authorize endpoint, with the redirect URI "https://api.example.com" and state "sample-state"
+    Then the user will be redirected to the confirm details page
+    When the user clicks Continue
+    Then the user will be redirected to the client's redirect URI with an authorization code
```
michaelabenyohai: I forgot to write in the story, but can we also test that the `state` is returned here?

So the step will become:

`Then the user will be redirected to the client's redirect URI with an authorization code and the state`

Note that we we will test that the value of the state is the same as we put in the first step, so `sample-state`.

The `state` is in the query params on the redirect, like code is.

Per https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2

```
HTTP/1.1 302 Found
Location: https://client.example.com/cb?code=SplxlOBeZQQYbYS6WxSbIA
          &state=xyz
```

tests/acceptance/steps/utils/auth-api.ts
```diff
@@ -94,3 +94,25 @@ export const token = async (parameters: TokenRequest): Promise<TokenResponse | O
...
+  const response = await request(publicApi)
+    .post("/confirm-details")
+    .set("content-type", "application/x-www-form-urlencoded")
+    .send({ redirectUri, state });
+
+  if (response.statusCode === 302) {
+    const { origin, search } = new URL(response.header["location"]);
+    return {
+      origin,
+      code: new URLSearchParams(search).get("code"),
```
michaelabenyohai, line 120: As per an earlier comment, this will also include `state` as a query param. Could you add that here. I think you'll also want to update the definition of the `authorizationResponse` type to be:

```
export type AuthorizationResponse = {
  origin: string;
  code: string | null;
  state: string | null;
};
```

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
-  function () {
-    const domainName = process.env.DOMAIN_NAME;
-    if ("origin" in this.authorizationResponse!) {
-      this.authorizationResponse.origin = domainName!;
-    }
+  "the user will be redirected to the client's redirect URI with an authorization code",
+  async function () {
     assert.ok(isAuthorizationResponse(this.authorizationResponse));
-    assert.equal(this.authorizationResponse.origin, domainName);
+    assert.equal(this.authorizationResponse.origin, "https://api.example.com");
```
michaelabenyohai: Given that on line 24 you set the value of `this.redirectUri`, I think it might make sense to change this to:

`assert.equal(this.authorizationResponse.origin, this.redirectUri);`

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
 });
 
 When<WorldDefinition>(
-  "I call the authorize endpoint, with the redirect URI {string}",
-  async function (redirectUri: string) {
+  "the client calls the authorize endpoint, with the redirect URI {string} and state {string}",
+  async function (redirectUri: string, state: string) {
+    this.redirectUri = redirectUri;
+    this.state = state;
     this.authorizationResponse = await authorize({
```
michaelabenyohai: Technically, in OAuth terms - the response from `/authorize` is not the "authorization response". It used to be, which is why this code is written like this.

But at some point we introduced the `confirm-details` page and now the "authorization response" (the official OAuth one) comes after we call `POST /confirm-details`.

So I would advocate that we don't set the response from `GET /authorize` to `this.authorizationResonse`.

Maybe we have a new property in the `WorldDefinition`, but I'm wondering what to call these things. Maybe it should look something like this:?

```
export type WorldDefinition = {
  ...
  userIdentityPostResponse?: Response;
  getAuthorizeResponse?: RedirectResponse; # <- set this here
  postConfirmDetailsResponse?: AuthorizationResponse | OAuthBadRequest; # <- this is the old authorizationResponse, now to be set from call to POST /confirm-details
  ...
};

type RedirectResponse = {
  origin: string;
}
```

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
+    assert.ok(this.authorizationResponse.code, "Expected an authorization code but got none");
   }
 );
 
-When<WorldDefinition>("I call the token endpoint with the authorization code", async function () {
+When<WorldDefinition>("the client calls the token endpoint with the authorization code", async function () {
   assert.ok(isAuthorizationResponse(this.authorizationResponse));
   this.tokenResponse = await token({
     grant_type: TokenGrantType.AuthorizationCode,
     code: "SplxlOBeZQQYbYS6WxSbIA",
```
michaelabenyohai: Now that we have the authorization code being returned in the redirect to the redirect URI (your new step!), we can stop hardcoding this value. We can effectively [revert this line of code](https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/367/changes/d0a52d65c9ad1c7bd23764d3385e7dc288015ad0#diff-6dee371ac3ec66767beb76079722bd4999f01b405582ce6f7ea0eb21876882acL51) (which we had to hack to make the tests pass).

So this becomes `code: this.authorizationResponse.code`

tests/acceptance/steps/utils/auth-api.ts
michaelabenyohai: Note that the response to `GET /authorize` is a redirect to `GET /confirm-details` but it would not include any query parameters. We can remove this line of code (77).

The `code` now is returned from a call to `POST /confirm-details`, which you already do in your new function on line 113.

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
-    if ("origin" in this.authorizationResponse!) {
-      this.authorizationResponse.origin = domainName!;
-    }
+  "the user will be redirected to the client's redirect URI with an authorization code",
+  async function () {
     assert.ok(isAuthorizationResponse(this.authorizationResponse));
-    assert.equal(this.authorizationResponse.origin, domainName);
+    assert.equal(this.authorizationResponse.origin, "https://api.example.com");
+    assert.ok(this.authorizationResponse.code, "Expected an authorization code but got none");
   }
```
michaelabenyohai, line 57: As per another comment, can we also check here that the `state` returned from `POST /confirm-details` is the same value as the one provided on `GET /authorize`.

```
assert.equal(this.authorizationResponse.state, this.state);
```

tests/acceptance/steps/base-verbs.step.ts
```diff
@@ -15,6 +15,8 @@ export type WorldDefinition = {
   userIdentityPostResponse?: Response;
   authorizationResponse?: AuthorizationResponse | OAuthBadRequest;
   tokenResponse?: TokenResponse | OAuthBadRequest;
+  redirectUri?: string;
+  state?: string;
```
michaelabenyohai: Can you move these two values further up this list, above the `*Response` properties? It's a small thing, but these are things the test cases set early on, and the responses are things that are set later on, in scenarios.

So in my head it makes sense to have these things further up the list to help with understanding how these tests work.

tests/acceptance/steps/base-verbs.step.ts
```diff
@@ -34,6 +36,8 @@ Before<WorldDefinition>(async function () {
   this.keyId = await getSigningKeyId();
   this.authorizationResponse = undefined;
   this.tokenResponse = undefined;
+  this.redirectUri = undefined;
+  this.state = undefined;
```
michaelabenyohai: Same as above, maybe move these further up?

tests/acceptance/steps/authorization.step.ts
```diff
@@ -11,51 +12,63 @@ import {
...
+  const domainName = process.env.DOMAIN_NAME;
+  if ("origin" in this.authorizationResponse!) {
+    this.authorizationResponse.origin = domainName!;
+  }
+  assert.ok(isAuthorizationResponse(this.authorizationResponse));
+  assert.equal(this.authorizationResponse.origin, domainName);
+});
+
+When<WorldDefinition>("the user clicks Continue", async function () {
+  this.authorizationResponse = await confirmDetailsSubmission(this.redirectUri!, this.state!);
```
michaelabenyohai, line 47: This makes complete sense.

There is one downside that if we change/break the button such that it no longer calls this endpoint, the acceptance tests would still pass.

But the other way of doing this would be to interact with the HTML on the page, which feels like it could be a separate suite of tests.

@chrisclayson any thoughts on this?

joe-mcgarry: Yeah, I figured I'd go with this approach first because the only other way I have a rough idea about is using Playwright to get elements on the page - but for 1 button, it felt like overkill. But if it's something we'd want, to align with other CRI teams who have browser tests or simply for our own reassurance, then I can look into doing it (unless there's another way I don't know about)

deploy-to-dev.sh
```diff
@@ -165,7 +165,7 @@ if $DEPLOY_SIS; then
 
     if $DEPLOY_OAUTH; then
         OAUTH_STACK_NAME="$STACK_NAME-oauth-internal"
-    elif
+    else
```
joe-mcgarry, line 168: There is now, but it wasn't when this PR was initially raised (as a bit of context around why it's included in this rather than not simply rebasing main as you say)

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: I'm of the same mind that it shouldn't be needed. What I can say as of this moment is that without the if statement, the code doesn't think the authorizationResponse has an origin property so gives a nice little error. I'm sure there's an elegant way around it, but it is a job for tomorrow's brain - but wanted to at least give a mini explanation of why it was likely there now/in the first place

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: I've created `RedirectResponse` and I think I've updated where it's needed in order to better reflect the OAuth flow as you said. Let me know what you think

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: Done!

tests/acceptance/steps/utils/auth-api.ts
joe-mcgarry: I've added those in now

tests/acceptance/features/authorization-workflow.feature
joe-mcgarry: I think I've made this change

tests/acceptance/features/authorization-workflow.feature
joe-mcgarry: Done!

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: Good spot, I've change it now

tests/acceptance/steps/utils/auth-api.ts
joe-mcgarry: I've amended this now

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: Done!

tests/acceptance/steps/base-verbs.step.ts
joe-mcgarry: Sure, I've moved those (and for the comment below too) higher up. Let me know if you'd prefer them in a different position

tests/acceptance/steps/base-verbs.step.ts
joe-mcgarry: Done - let me know if you're happy with where they are now

tests/acceptance/steps/utils/auth-api.ts
```diff
@@ -57,12 +62,13 @@ export const isAwsError = (response: unknown): response is AwsError => {
   return !!response && typeof response === "object" && "message" in response;
 };
 
+export const isRedirectResponse = (response: unknown): response is RedirectResponse =>
+  !!response && typeof response === "object" && "origin" in response;
+
 export const isAuthorizationResponse = (response: unknown): response is AuthorizationResponse =>
   !!response && typeof response === "object" && "origin" in response && "code" in response;
```
michaelabenyohai: I've just noticed that checking for `state` is missing in this type guard now. Can you add it?

tests/acceptance/steps/authorization.step.ts
michaelabenyohai: Yeah I agree, this is good and we can have a conversation about using PlayWright separately.

tests/acceptance/steps/utils/auth-api.ts
joe-mcgarry: Wil do!

tests/acceptance/steps/base-verbs.step.ts
```diff
@@ -2,17 +2,20 @@ import { setDefaultTimeout, Before, defineParameterType } from "@cucumber/cucumb
...
   keyId: string;
   userId: string;
   bearerToken?: string;
+  redirectUri?: string;
+  state?: string;
   requestedVtr: string[];
   govukSigninJourneyId: string;
   credentialJwts: string[];
   userIdentityPostResponse?: Response;
+  redirectResponse?: RedirectResponse | OAuthBadRequest;
```
michaelabenyohai: We can remove the `| OauthBadRequest` from this line. This would never get set to that - it would happen later when we set `authorizationResponse`

src/identity-reuse/verifiable-credential-jwt.ts
```diff
@@ -5,7 +5,9 @@ import {
 } from "@govuk-one-login/data-vocab/credentials";
 
 export type VerifiableCredentialJWT =
-  IdentityCheckCredentialJWTClass | RiskAssessmentCredentialJWTClass | SecurityCheckCredentialJWTClass;
+  | IdentityCheckCredentialJWTClass
```
michaelabenyohai: I'm not sure if this was intentional, but looks like the additional `|` is causing linting issues.

tests/acceptance/steps/authorization.step.ts
```diff
@@ -33,11 +35,9 @@ When<WorldDefinition>(
...
-Then<WorldDefinition>("the user will be redirected to the confirm details page", function () {
-  const domainName = process.env.DOMAIN_NAME;
-  if ("origin" in this.redirectResponse!) {
-    this.redirectResponse.origin = domainName!;
-  }
+Then<WorldDefinition>("the user will be redirected to the confirm details page", async function () {
+  const domainName = await getCloudFormationOutput(CloudFormationOutputs.SisPublicApi);
+  assertRedirectResponse(this.redirectResponse);
   assert.ok(isRedirectResponse(this.redirectResponse));
   assert.equal(this.redirectResponse.origin, domainName);
```
michaelabenyohai, line 42: This looks good. I've just realised one thing. The description says "the user will be redirected to the confirm details page" but we're not actually checking that they are at that page (just the right domain).

I think we need a `assert.equal(this.redirectResponse.pathname, "/confirm-details");`

In `auth-api.ts` I think we can do the following to get this:

```
  if (response.statusCode === 302) {
    const { origin, pathname } = new URL(response.header["location"]);
    return {
      origin,
      pathname,
    };
  }
```

src/identity-reuse/verifiable-credential-jwt.ts
joe-mcgarry: Believe it or not, this was the linter which complained and wanted to change it locally. Now the GitHub cousin is saying to change it back

tests/acceptance/steps/authorization.step.ts
joe-mcgarry: Good catch! So we want the domain name and a new assertion for the path? Or just the path name?

michaelabenyohai: I think in this case `path` and `pathname` are the same thing? I would call it `path` but the URL type calls it `pathname` I think. Whichever you think to check it goes to the right page.

tests/acceptance/steps/base-verbs.step.ts
michaelabenyohai: If I'm interpreting the original code correctly, I think `OAuthBasRequest` might be trying to reflect the possibility of getting an [error response](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1) from the OAuth authorization endpoint instead of a [successful authorization response](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2).

At the moment that would come from `/confirm-details`. But having thought about, in future maybe it will be possible for it to come from this endpoint. It's doing no harm so maybe just leave it as it is for now.

joe-mcgarry: I do agree with your interpretation! Maybe another error type (or a generic Error) could be made to go here, because presumably we'd want to check for when `if it isn't a 302` scenarios? I don't think I worded it well at the end of the day in my last comment (and definitely mixed up the 2 response 🙃 ) but I can add some sort of error, or leave it as is if you'd prefer?

michaelabenyohai (approved):

PR #375, opened 2026-07-07
SPT-1969: Don't deploy phase 3 resources to prod yet

michaelabenyohai:
## What

Don't deploy the `UserIdentityHandlerFunction` and various log groups to integration of prod.

## Why

The UserIdentityHandlerFunction is not being used in production yet and we want to only deploy it up to staging whilst we are still building the OAuth server. The same is true for various log groups associated with the build of the OAuth server.

### DIFF
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -517,6 +517,7 @@ Resources:
 
   AuthorizationFunctionLogGroup:
     Type: AWS::Logs::LogGroup
+    Condition: IsNotProduction
     Properties:
       LogGroupName: !Sub "/aws/lambda/${AWS::StackName}-AuthorizationFunction"
       RetentionInDays: 30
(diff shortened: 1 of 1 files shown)


umaar (approved):

mislam246 (approved):

chrisclayson (approved):

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -750,6 +752,7 @@ Resources:
   ################################
   UserIdentityHandlerFunction:
     Type: AWS::Serverless::Function
+    Condition: IsNotProduction
     Properties:
       FunctionName: !Sub "${AWS::StackName}-UserIdentityHandlerFunction"
       Description: "A lambda function that queues SIS data for processing"
```
chrisclayson, line 758: I know not part of this particular change but this isn't what this function does 🙈

michaelabenyohai: I'll update this in my next PRs which do the renames - I've also been updating some other nonsensical descriptions there

PR #376, opened 2026-07-07
BAU: Consistent bash quoting and bracing in deploy script

chrisclayson:
## What

- Replace rogue `elif` with `else`
- Apply consistent quoting and curly brackets around variable in `deploy-to-dev.sh` script.
- Use of array variables and expansion where passing multiple arguments to be interpolated to the command line.

## Why

The `elif` was causing an error while execting.

The array values refactor will correctly allow multiple values to be passed onto to the `deploy_or_destroy` function, this in turn will allow multiple "parameter overrides" to safely be passed.

The `${envvar}` vs `$envvar` fixes are purely for consistency.

### DIFF
--- a/deploy-to-dev.sh
+++ b/deploy-to-dev.sh
@@ -2,6 +2,7 @@
 
 set -e
 set -o pipefail
+BASE_DIR="$(dirname "${0}")"
 
 usage() {
   cat << EOF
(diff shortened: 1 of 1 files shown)


deploy-to-dev.sh
```diff
@@ -140,57 +141,56 @@ while [[ -n "$1" ]]; do
...
+if "${DEPLOY_SIS}"; then
+    SIS_STACK_NAME="${STACK_NAME}"
+    if "${DEPLOY_OAUTH}"; then
+        OAUTH_STACK_NAME="${STACK_NAME}-oauth"
+    else
         OAUTH_STACK_NAME="preview-main-oauth"
     fi
-elif $DEPLOY_OAUTH; then
-  OAUTH_STACK_NAME=$STACK_NAME
+elif "${DEPLOY_OAUTH}"; then
```
michaelabenyohai, line 171: Do you want this to be `else`?

michaelabenyohai (approved):

deploy-to-dev.sh
chrisclayson: No, this one is fine as it has an associated condition. The one above a pure `else` scenario (like the one on line 173).

PR #377, opened 2026-07-07
SPT-2005: Update stacks to use TxMA queue from shared stack

chrisclayson:
## What

Update the OAuth internal and main SIS stack stacks to include a new parameter (`SharedStackName`) that contains the name of the shared stack to use for the deployment.

Use the exports from the specified shared stack to define the TxMA audit queues the functions use and the appropriate permissions.

## Why

This removes a circular dependency between the two stacks where the OAuth stack needs SIS for the TxMA queues, but SIS needs the OAuth stack for API Gateway URL.

## Related PRs

https://github.com/govuk-one-login/ipv-identity-reuse-service-config/pull/159
https://github.com/govuk-one-login/ipv-identity-reuse-service-config/pull/163

### DIFF
--- a/.github/workflows/deploy-to-aws.yml
+++ b/.github/workflows/deploy-to-aws.yml
@@ -14,6 +14,12 @@ on:
         default: preview-pr-${{ github.event.pull_request.number }}-oauth"
         required: false
         type: string
+      capabilities:
+        description: >
+          The capability list passed to CloudFormation (normally, the default is sufficient)
+        default: 'CAPABILITY_NAMED_IAM'
+        required: false
+        type: string
       environment-name:
         description: >
           The name of the environment. If this is not provided it's taken from the branch name.
@@ -110,12 +116,10 @@ jobs:
         env:
           STACK_NAME: ${{ inputs.stack-name }}
           ENVIRONMENT_NAME: ${{ inputs.environment-name }}
-          OAUTH_STACK_NAME: ${{ inputs.oauth-stack-name }}
         run: |
           echo "ENVIRONMENT_NAME=${ENVIRONMENT_NAME}" >> "$GITHUB_ENV"
           echo "STACK_NAME=${STACK_NAME}" >> "$GITHUB_ENV"
           echo "S3_PREFIX=${STACK_NAME}" >> "$GITHUB_ENV"
-          echo "OAUTH_STACK_NAME=${OAUTH_STACK_NAME}" >> "$GITHUB_ENV"
 
       - name: Delete failed stack
         uses: govuk-one-login/github-actions/sam/delete-stacks@0eb1972f8c0d539b3ff4c24d78b3dba917343e7c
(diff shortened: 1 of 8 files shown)


deploy-to-dev.sh
```diff
@@ -184,13 +183,14 @@ fi
 export AWS_DEFAULT_REGION=eu-west-2
 echo "Environment: ${ENVIRONMENT}"
 echo "Profile:     ${AWS_PROFILE}"
-$DEPLOY_SIS && echo "Deploy identity-reuse-service to stack ${SIS_STACK_NAME}"
-$DEPLOY_OAUTH && echo "Deploy oauth-internal to stack ${OAUTH_STACK_NAME}, pointing at identity-reuse-service stack ${SIS_STACK_NAME}"
+$DEPLOY_SIS && echo "Deploy identity-reuse-service to stack ${SIS_STACK_NAME}, pointing at shared stack ${SHARED_STACK_NAME}"
```
joe-mcgarry, line 186: With SIS_STACK_NAME being defined/set on line 164 inside an if statement, I wanted to double check if that would keep its scope to the if statement only or if I'm just being a pedant and things are all good in bash world with it being mentioned here (and on line 195)

test.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+BOOLEAN=false
```
joe-mcgarry: Double checking if this file is needed?

.github/workflows/deploy-to-aws.yml
```diff
@@ -14,6 +14,12 @@ on:
         default: preview-pr-${{ github.event.pull_request.number }}-oauth"
         required: false
         type: string
+      capabilities:
+        description: >
+          The capability list passed to CloudFormation (normally, the default is sufficient)
+        default: 'CAPABILITY_NAMED_IAM'
```
joe-mcgarry, line 20: Is this now being pass as an input because the main SIS stack only needs CAPABILITY_NAMED_IAM but the Oauth stack needs this and the AUTO_EXPAND capability?

.github/workflows/deploy-to-aws.yml
```diff
@@ -124,24 +128,11 @@ jobs:
...
-      #          sam deploy \
-      #          --stack-name "$OAUTH_STACK_NAME"
-      #          --region "$AWS_REGION"  \
-      #          --no-confirm-changeset \
-      #          --no-fail-on-empty-changeset \
-      #          --capabilities CAPABILITY_NAMED_IAM,CAPABILITY_AUTO_EXPAND \
-      #          --tags DeploymentSource="GitHub Actions" StackType=Preview \
-      #          --parameter-overrides Environment="dev" SisStackName=preview-main
-
       - name: Deploy SAM Application
```
joe-mcgarry, line 131: Is the idea with this workflow/step to only deploy the main SIS stack and not have it deploy the OAuth one (due to that not really changing outside of specific, known to us changes) so we just need to point it in the right direction?

joe-mcgarry:
Not entirely sure, but would we need to pass in the OAuth stack name and any capabilities in the post-merge workflow at all when it calls `deploy-to-aws` too?

test.sh
chrisclayson: Doh ... no! You've just had sneak preview of something I was going to talk about at the next tech time 😂

.github/workflows/deploy-to-aws.yml
chrisclayson: Correct!

.github/workflows/deploy-to-aws.yml
chrisclayson: I've refactored to make `deploy-to-aws` generic and can be use to deploy either stack. If you look at the changes to `pre-merge` workflow, you'll see I call `deploy-to-aws` twice, once for the OAuth stack and once for the SIS stack.

chrisclayson: Mainly did this to separate the usages of the build artifacts it downloads before deploying.

chrisclayson:
> Not entirely sure, but would we need to pass in the OAuth stack name and any capabilities in the post-merge workflow at all when it calls `deploy-to-aws` too?

Oh, good point, you're right ... I missed that 🤦🏻

deploy-to-dev.sh
chrisclayson: The `$DEPLOY_SIS &&` at the beginning of this line is essentially doing the same `if` statement, so this line will only echo if `DEPLOY_SIS` is true.

chrisclayson:
`post-merge` update now pushed as a new commit.

.github/workflows/deploy-to-aws.yml
chrisclayson: Will also make it easier to selectively deploy the stacks based on what has changed ... if we choose to do that in the future 🤓

deploy-to-dev.sh
joe-mcgarry: My question was more about the scope of SIS_STACK_NAME, with it now only being declared and defined inside an if statement (thinking it would be like other languages where it isn't accessible/existing outside of the if block), but I did a little test and with your comment now know for sure that SIS_STACK_NAME does indeed get set

.github/workflows/deploy-to-aws.yml
joe-mcgarry: Makes sense!

joe-mcgarry (approved):

PR #379, opened 2026-07-09
SPT-1973: Fix error when reading API response

mislam246:
## What
Change the logic to read the `/api/authorization` endpoint's response body once, instead of re-reading it for both the Type Guard function and accessing the properties for the redirect.

## Why
To resolve the error caused by consuming the data stream multiple times.

## Related PRs
https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/364
https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/378

### DIFF
--- a/src/handlers/get-callback-handler/get-callback-handler.ts
+++ b/src/handlers/get-callback-handler/get-callback-handler.ts
@@ -52,12 +52,13 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
 };
 
 const redirectToClient = async (authorizationResponse: Response, domainName: string) => {
-  if (!isValidSuccessResponse(await authorizationResponse.json())) {
+  const authResponseJson = await authorizationResponse.json();
+  if (!isValidSuccessResponse(authResponseJson)) {
     logger.error("Invalid response properties received from authorization endpoint");
     return await redirectToErrorPage(domainName);
   }
 
-  const authorizationData = (await authorizationResponse.json()) as AuthorizationSuccessResponse;
+  const authorizationData = authResponseJson as AuthorizationSuccessResponse;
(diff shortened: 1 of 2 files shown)


src/handlers/get-callback-handler/get-callback-handler.ts
```diff
@@ -52,12 +52,13 @@ export const handler = async (event: APIGatewayProxyEvent): Promise<APIGatewayPr
...
 const redirectToClient = async (authorizationResponse: Response, domainName: string) => {
-  if (!isValidSuccessResponse(await authorizationResponse.json())) {
+  const authResponseJson = await authorizationResponse.json();
+  if (!isValidSuccessResponse(authResponseJson)) {
     logger.error("Invalid response properties received from authorization endpoint");
     return await redirectToErrorPage(domainName);
   }
 
-  const authorizationData = (await authorizationResponse.json()) as AuthorizationSuccessResponse;
+  const authorizationData = authResponseJson as AuthorizationSuccessResponse;
```
michaelabenyohai, line 61: Do we have unit test that exercises this code? I am wondering why the unit tests didn't catch this problem. Can you check, and add one?

mislam246: I've added a test to check that the response body is read only once.

From what I've read:
It wasn't catching the problem in the original unit tests because the `mockResolvedValue()` returns a plain JS object and not a real Response object, so it wouldn't throw an error if the body is read multiple times. However, when testing it in dev, the returned Response uses a Stream, so can only be called once (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#locked_and_disturbed_streams).

michaelabenyohai (approved):

PR #382, opened 2026-07-10
SPT-1969: Rename production lambdas - part 1

michaelabenyohai:
## What

Rename some lambdas to make it more obvious what they do:

* `UserIdentityFunction` to `PostPhase2UserIdentityHandler` (this one will be deprecated in future)
* `MessageProcessorLambda` to `InboundTxmaQueueMessageHandler` 

The metrics namespaces and services have kept the same name for now. A future PR will improve the way we do this.

Also fix blue/green deploys of the `MessageProcessorambda`.

> [!NOTE]
>  This PR duplicates the `UserIdentityFunction` and `MessageProcessorLambda` CloudFormation resources with their respective new name but does not yet switch the API or EventSourceMapping over to using the new versions. To ensure this deploys in a zero-downtime manner a future PR will switch the API and EventSourceMapping over, then tidy up the old resources.

## Why

We had lots of similarly named Lambdas.

## Testing

Passing acceptance tests should mean the `UserIdentityFunction` rename is ok.

We test the `MessageProcessorLambda` rename manually:

Use the following message:

```
{
  "user_id": "test-user-id",
  "timestamp": "1783691629",
  "intervention_code": "12"
}
```

Go to SQS -> preview-pr-382-stub-queue -> Send and receive messages -> Put above message into message body and send

Go to Lambda -> preview-pr-382-MessageProcessorLambda -> Monitor -> check you see a successful invocation

## Related PRs

* https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/363


openAPI/api.yaml
```diff
@@ -58,9 +58,9 @@ paths:
       x-amazon-apigateway-integration:
         httpMethod: POST
         uri:
-          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${UserIdentityFunction.Arn}:live/invocations
+          Fn::Sub: arn:aws:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/${PostPhase2UserIdentityHandler.Arn}:live/invocations
         credentials:
-          Fn::GetAtt: UserIdentityRestApiRole.Arn
+          Fn::GetAtt: PostPhase2UserIdentityHandlerRestApiRole.Arn
```
chrisclayson: This is the bit I'm nervous about ... I think, I would do this in the next PR (and before the PR that removes the old versions).

michaelabenyohai: Yes that was unintentional - I've removed it.

chrisclayson (dismissed):

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -2211,6 +2366,21 @@ Resources:
       Roles:
         - !Ref UserIdentityRestApiRole
 
+  PostPhase2UserIdentityHandlerRestApiPolicy:
+    Type: AWS::IAM::Policy
+    Properties:
+      PolicyName: !Sub "${AWS::StackName}-UserIdentityRestApiPolicy"
```
michaelabenyohai: I just noticed this is wrong, changing

chrisclayson (approved):

PR #393, opened 2026-07-14
SPT-2019: Harden Audit Queue Permissions

joe-mcgarry:
## What

Refactors the `sqs:SendMessage` policy for lambdas relying on sending messages to the Shared Stack's `AuditEventQueue`
The new policy is imported to the lambda role's `ManagedPolicies` property, with the permission in the lambda policy being deleted

## Why

There was lots of duplication for setting this policy, so it's been refactored to use a now exported policy (which grants `sqs:SendMessage` to the same queue) rather than repeatedly defining that policy in this template. 

## additional sections as needed

The config shared stack and then an SIS stack pointing to the recently deployed shared stack were deployed successfully

## Related PRs

The shared stack changes: https://github.com/govuk-one-login/ipv-identity-reuse-service-config/pull/168

### DIFF
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -2019,6 +2019,7 @@ Resources:
         - arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess
         - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
         - Fn::ImportValue: !Sub "${AppConfigStackName}-AppConfigClientPolicyArn"
+        - Fn::ImportValue: !Sub "${SharedStackName}-AuditEventQueueSendMessagePolicy"
       Tags:
         - Key: Product
           Value: !Ref ProductTagValue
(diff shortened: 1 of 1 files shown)


infrastructure/identity-reuse-service/template.yaml
```diff
@@ -2019,6 +2019,7 @@ Resources:
         - arn:aws:iam::aws:policy/AWSXrayWriteOnlyAccess
         - arn:aws:iam::aws:policy/service-role/AWSLambdaVPCAccessExecutionRole
         - Fn::ImportValue: !Sub "${AppConfigStackName}-AppConfigClientPolicyArn"
+        - Fn::ImportValue: !Sub "${SharedStackName}-AuditEventQueueMessagePolicy"
```
joe-mcgarry: Is there a better way to go about importing values? I couldn't find a nice way from searching, besides `Fn::GetStackOutput` but that doesn't solve the entire problem of repeating this policy import (and I didn't think adding some sort of parameter higher up would work)?

chrisclayson: This is fine, we do this a lot (see the line above). There isn't really an alternative.

PR #396, opened 2026-07-15
SPT-2007: Refactor and remove duplicated code

mislam246:
## What
Refactor handlers by separating out the OAuth Internal API calls and redirect logic into a new directory. Did not include the other handlers since they either have front-end renders or return a response without redirecting to another URL.

## Why
To reduce duplication and improve reusability, particularly in the `get-authorize-handler` and `get-callback-handler` files.


src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,49 @@
...
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+    signal: AbortSignal.timeout(SESSION_TIMEOUT_MS),
+  });
+}
+
+export async function getAuthorizationCode(
+  queryParameters: APIGatewayProxyEventQueryStringParameters,
```
michaelabenyohai: It would be nice to make this layer completely independent of the Handler and API, and ensure it is not dependent on the shape of the API.

At the moment, given it is consuming `APIGatewayProxyEventQueryStringParameters`, it would need to be changed if the SIS API changed.

Can you instead make the function declaration:

```
export async function getAuthorizationCode(
  clientId: string, redirectUri: string, state: string)
```

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,49 @@
...
+  sessionId: string
+): Promise<Response> {
```
michaelabenyohai: Sorry for not putting this in the story, but one of the aims of this separation is to keep knowledge of the various APIs in one place. By this I mean it would be good to keep knowledge of the OAuth internal API in this file only, and ensure that we don't "leak" awareness of the OAuth internal API's behaviour (via response codes) into the Handlers above. If the OAuth internal API changes, we should only have to update this file, and not the Handlers.

Therefore it would be preferable if this returned a new type representing the Authorization Code (or error) rather than the HTTP `Response` object. We should do the error/status handling for the `Response` in this function rather than the handler.


I think this should return something like:

```
Promise<AuthorizationResult>
```

where for `200` or `403` we return:

```
type AuthorizationResult {
  authorizationCode?: string,
  redirectUri: string
  state: string
}
```

and if not `200` or `403` we throw an error that is caught and handled in the Handler above.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,49 @@
...
+  const oauthInternalApiUrl = process.env.OAUTH_INTERNAL_API;
```
michaelabenyohai: We seem to have two different ways of getting the URL for OAuth-internal in this file - here and in the other function, above. Can you make them consistent by pulling them into a single private function in this file?

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type SessionResponse = {
+  session_id: string;
+  state: string;
+  redirect_uri: string;
+};
+
+export async function callSessionApi(clientId: string, request: string): Promise<Response> {
+  const oauthInternalApiUrl = getRequiredEnvironment("OAUTH_INTERNAL_API_URL");
+  const SESSION_TIMEOUT_MS = 10_000;
```
michaelabenyohai: I know you didn't add this, but putting these together has made it obvious that we use a timeout in one and not the other. 10 seconds sounds long so we might want to remove this. I'll think about it and let you know.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,49 @@
...
+import { APIGatewayProxyEventQueryStringParameters } from "aws-lambda";
+import { getRequiredEnvironment } from "../commons/get-required-environment";
+
+export type SessionResponse = {
+  session_id: string;
+  state: string;
+  redirect_uri: string;
+};
+
+export async function callSessionApi(clientId: string, request: string): Promise<Response> {
```
michaelabenyohai: As with the other function, can we have this return a type representing the information coming back from the API, rather than the `Response` itself, which leaks logic into the handler.

I imagine this one will just be:

```
Promise<SessionResult>
```

where:

```
type SessionResult {
  sessionId: string,
  redirectUri: string,
  state: state,
}
```

Note that this is exactly the same as the `SessionResponse` object in this file, so I would rename `SessionResponse` to `SessionResult` (latter sounds a bit less API-specific).

You will need to do the above for `201` status codes. And anything else throw an error, which can be caught and handled in the Handler code. You might want to consider throwing a custom error named something like `CreateSessionError`.

src/handlers/get-authorize-handler/get-authorize-handler.ts
```diff
@@ -85,39 +30,20 @@ export async function handler(event: APIGatewayProxyEvent, context: Context): Pr
...
-    logger.error(`Error calling session handler: ${error}`);
-    return redirectToError(domainName);
+    let response: Response;
+    try {
+      response = await callSessionApi(clientId, request);
+    } catch (error) {
+      logger.error(`Error calling session handler: ${error}`);
+      return redirectToErrorPage(domainName);
+    }
+    if (response.status !== 201) {
```
michaelabenyohai: As per other comments, I think this will become catching an error thrown from `callSessionAPI`, which can likely be incorporated into the code above.

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function redirectToClientWithError(redirectUri: string, state: string) {
+  const orchestrationRedirectUrl = new URL(redirectUri);
+  orchestrationRedirectUrl.searchParams.append("error", "access_denied");
+  // To be replaced with the state returned by the /authorization endpoint once it's added to the oauth-common response
+  orchestrationRedirectUrl.searchParams.append("state", state);
+  return redirect(`${orchestrationRedirectUrl}`, "");
+}
+
+export async function redirect(location: string, body: string, includeCookie = false, sessionId?: string) {
```
michaelabenyohai: I think this should have no knowledge of `sessionId`. Instead I think this should just take an optional `cookie`.

```
export async function redirect(location: string, body: string, cookie?: string)
```

Then you can use `(cookie && { "Set-Cookie": cookie })`

`buildSessionCookie` can be called outside of this function and the result passed in.

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,68 @@
...
+  const value = `${SESSION_COOKIE_NAME}=${sessionId}`;
+  const attributes = "Path=/; Secure; HttpOnly; SameSite=Lax";
+  return `${value}; ${attributes}`;
+}
+
+export async function redirectToErrorPage(domainName: string) {
+  return await redirect(`https://${domainName}/error/unrecoverable`, "");
+}
+
+export async function redirectToConfirmDetailsWithCookie(domainName: string, sessionResponse: SessionResponse) {
```
michaelabenyohai: This function is almost identical to `redirectToConfirmDetails` and duplicates quite a lot of code that it doesn't need to. I think you can completely remove this and just have one method for `redirectToConfirmDetails`

```
export async function redirectToConfirmDetails(domainName: string, state: string, redirectUri: string, cookie?: string) {
  const url = new URL("/confirm-details", `https://${domainName}`);
  url.searchParams.append("state", state);
  url.searchParams.append("redirect_uri", redirectUri);

  return await redirect(url.href, "", cookie);
}
```

I think it is fine to leave the `buildSessionCookie` method in the handler that uses it and pass the result in if it is required.

src/services/oauth-internal-service.ts
umaar: Had originally put that there to have something rather than nothing!

One suggestion is to use timeouts in both places: but will defer to Michaela

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
```diff
@@ -1,20 +1,24 @@
...
+import { afterAll, beforeEach, describe, expect, it, MockInstance, vi, vitest } from "vitest";
 import { AuthorizationQueryStringParameters, handler } from "../get-authorize-handler";
+import * as oauthInternalService from "../../../services/oauth-internal-service";
 
 process.env.DOMAIN_NAME = "test-domain";
 process.env.OAUTH_INTERNAL_API_URL = "https://example.com/v1";
 
-const mockFetch = vi.fn();
-vi.stubGlobal("fetch", mockFetch);
+vi.mock("../../../services/oauth-internal", () => ({
```
umaar: I think `oauth-internal` doesn't exist now. I guess it's `oauth-internal-service`, curious why vitest didn't complain! Also then it raises a question about whether the `unstubAllGlobals` is actually needed?

In addition to fixing, might be worth considering (in the future) why Vitest didn't complain just in case we're suppressing important information.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
```diff
@@ -87,17 +92,7 @@ describe("authorize-handler", () => {
...
-      expect(mockFetch).toHaveBeenCalledWith("https://example.com/v1/api/session", {
-        method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-        },
-        body: JSON.stringify({
-          client_id: "orchestrator",
-          request: "foo.bar.123",
-        }),
-        signal: expect.any(AbortSignal),
```
umaar, line 99: In the old code, we were also checking the abort signal was passed in, but now in the new code, I think we are not. There is that outstanding comment thread suggesting it's removed though.

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,77 @@
...
+beforeEach(() => {
+  vitest.clearAllMocks();
+});
+
+afterEach(() => {
+  vitest.restoreAllMocks();
+});
+
+it("should call the /api/session fetch and return the raw response", async () => {
+  process.env.OAUTH_INTERNAL_API_URL = "https://test.com";
```
umaar: Silly question but does this leak into other tests. Maybe `stubEnv/unstubAllEnvs` is better

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  const mockResponse = Response.json(
+    { state: "test-state", redirect_uri: "https://test-uri.com", session_id: "test-session-id" },
+    {
+      status: 200,
+      headers: { "Content-Type": "application/json" },
+    }
+  );
+
+  vitest.stubGlobal("fetch", vitest.fn().mockResolvedValueOnce(mockResponse));
```
umaar, line 46: Similar to the comment above but would this leak into other tests? I think the `unstubAllGlobals` can help

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,68 @@
+import { SessionResponse } from "./oauth-internal-service";
+import { URL } from "node:url";
+import {
+  AuthorizationSuccessResponse,
```
umaar: Interesting as this service now depends on a handler. One thought could be that the direction should be that handlers depend on services, rather than a service depending on a handler. Maybe it's not the most serious thing, but I do wonder if patterns like this could then be copied in the future which could make maintaibibility challenging.

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,68 @@
...
+
+const SESSION_COOKIE_NAME = "identity_reuse_service_session";
+
+function buildSessionCookie(sessionId: string): string {
+  const value = `${SESSION_COOKIE_NAME}=${sessionId}`;
+  const attributes = "Path=/; Secure; HttpOnly; SameSite=Lax";
+  return `${value}; ${attributes}`;
+}
+
+export async function redirectToErrorPage(domainName: string) {
```
umaar: There are a few functions here marked as `async` that don't do any async things. Could consider removing the `async` keyword as it could indicate to readers that it relies on the network or some other async operation. Another example `redirectToConfirmDetails` has no async operations.

src/services/sis-redirect-service.ts
umaar: One perspective is that `redirect(url.href, "", true, sessionResponse.session_id)` is hard to read at the call site. And params like `true` or `""` can be challenging to digest. Also because `sessionId` is optional but asserted with `!`, I think `redirect(loc, "", true)` could lead to `identity_reuse_service_session=undefined`. Maybe consider an `options` object.

Interesting post: https://www.teemutaskula.com/blog/avoid-positional-args

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export async function redirectToConfirmDetails(domainName: string, state: string, redirectUri: string) {
+  const url = new URL("/confirm-details", `https://${domainName}`);
+  url.searchParams.append("state", state);
+  url.searchParams.append("redirect_uri", redirectUri);
+
+  return await redirect(url.href, "");
+}
+
+export async function redirectToClient(authorizationResponse: Response, domainName: string) {
```
michaelabenyohai: I think too much of the logic has been moved from the Handler into here. The Handler still needs to be the one deciding where to redirect to based on the AuthorizationResponse, but then it used this file just to DO the redirect (passing in appropriate params). So please can you move that code back into the handler?

I think you only need one `redirectToClient` for both the "error" and "success" scenarios and it can look like:

```
export async function redirectToOAuthClient(uri: string, state: string, code?: code) {
  const orchestrationRedirectUrl = new URL(uri);
  orchestrationRedirectUrl.searchParams.append("state", state);

  if (code) {
    orchestrationRedirectUrl.searchParams.append("code", code);
  } else {
    orchestrationRedirectUrl.searchParams.append("error", "access_denied");
  }

  return redirect(`${orchestrationRedirectUrl}`, "");
}

```

All other logic that exists in this function still needs to be in the handler above.

src/services/oauth-internal-service.ts
michaelabenyohai: Yes agree, let's use the timeout in both places. But we might want to reduce it slightly, I'll think about that!

src/services/sis-redirect-service.ts
michaelabenyohai: I agree, it would be good to make sure Handlers depend on Services, and not the other way around.

src/services/oauth-internal-service.ts
mislam246: Makes sense. I've changed the function signature as suggested, but also included the `sessionId` in the parameters, since the `/api/authorization` endpoint needs it.

src/services/oauth-internal-service.ts
mislam246: Updated as suggested so the function now returns only that type instead of the entire Response, where the returned properties are determined by the API call response's status code. I've also moved the `isValidSuccessResponse` function (the type guard that validates the response properties) from the `get-callback-handler.ts` into this function.

src/services/oauth-internal-service.ts
mislam246: Made them both the same by reusing the `getRequiredEnvironment()` function. Also renamed the environment variable here and in `template.yaml` to be consistent with the other function above.

src/services/sis-redirect-service.ts
mislam246: Done.

src/services/sis-redirect-service.ts
mislam246: Agreed. Refactored these into one function now, which also includes the cookie in the redirect if it exists. As you mentioned, the handler calls `buildSessionCookie` and passes the value to this function now.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
mislam246: I've changed it to `oauth-internal-service`. That's a good point about why there weren't errors when using the wrong name.

Also not sure if we need the `unstubAllGlobals`. I don't think it impacts the callSessionApi mock anyway and the tests work without it. I've removed it for now, and if you feel it's needed, we can add it back in.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
mislam246: Not sure what the best approach would be for this. Should we leave out the check on the AbortSignal given the suggestion you mentioned?

src/services/tests/oauth-internal-service.test.ts
mislam246: You're right, it could leak into the other test files. I've used `stubEnv` and `unstubAllEnvs` as suggested.

src/services/tests/oauth-internal-service.test.ts
mislam246: Added the `unstubAllGlobals` to the `afterEach()`.

src/services/sis-redirect-service.ts
mislam246: Makes sense. Some of the refactoring above has moved the response checks to within the `oauth-internal-service` now and I've moved the `get-callback-response` logic to a new `oauth-internal-service-response` file in the services folder. This should mean the service doesn't depend on the handler anymore but the other way.

src/services/sis-redirect-service.ts
mislam246: Yeah, none of these functions perform asynchronous logic. I've removed the `async` from all of them.

src/services/sis-redirect-service.ts
mislam246: Thanks for explaining and providing the link, I've changed it as suggested. It's a good refactor, makes it more readable and easier to handle optional parameters.

src/services/sis-redirect-service.ts
mislam246: Refactored these into one function and moved the `isValidSuccessResponse()` check to the `oauth-internal-service`. I added the `.sort()` function on the URL to order the `state` query parameter last to match the URL format specified in the original ticket. Not sure if that's necessary, so can be removed. Otherwise I think we'd need to set the `state` in the if/else.

src/services/oauth-internal-service.ts
mislam246: Refactored similar to the other function.

src/handlers/get-authorize-handler/get-authorize-handler.ts
mislam246: Removed the check on the status code so it now redirects depending on if an error was thrown.

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,52 @@
...
+  url.searchParams.append("redirect_uri", redirectUri);
+  url.searchParams.append("client_id", clientId);
+
+  return cookie
+    ? redirect({ location: url.href, body: "", cookie: cookie })
+    : redirect({ location: url.href, body: "" });
+}
+
+export function redirectToClient(redirectUri: string, state: string, authorizationCode?: string) {
+  const orchestrationRedirectUrl = new URL(decodeURIComponent(redirectUri));
```
umaar: If you trace through the code that leads up to this, it turns out a decode is happening twice (see `oauth-internal-service.ts`) on line 89. It is possible to get a malformed URL from decoding twice. So just doing the decode once is enough

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+  }
+}
+
+export async function getAuthorizationCode(
+  clientId: string,
+  redirectUri: string,
+  state: string,
+  sessionId: string
+): Promise<AuthorizationResult> {
+  const oauthInternalApiUrl = getRequiredEnvironment("OAUTH_INTERNAL_API_URL");
```
umaar: I think there is a discrepancy  with `OAUTH_INTERNAL_API` and `OAUTH_INTERNAL_API_URL`. I believe this would throw because the variable is missing?

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,52 @@
...
+  clientId: string,
+  cookie?: string
+) {
+  const url = new URL("/confirm-details", `https://${domainName}`);
+  url.searchParams.append("state", state);
+  url.searchParams.append("redirect_uri", redirectUri);
+  url.searchParams.append("client_id", clientId);
+
+  return cookie
+    ? redirect({ location: url.href, body: "", cookie: cookie })
```
umaar: nit: you can ignore, but I don't think this ternary is needed because the `redirect` function already handles an absent cookie.

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,52 @@
...
+export function redirectToErrorPage(domainName: string) {
+  return redirect({ location: `https://${domainName}/error/unrecoverable`, body: "" });
+}
+
+export function redirectToConfirmDetails(
+  domainName: string,
+  state: string,
+  redirectUri: string,
+  clientId: string,
+  cookie?: string
```
umaar: nit: You can defo ignore this, it's purely a nit and not important....but if you see this comment: https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3613385046

I wonder if there's a similar point here, maybe 5 positional arguments is a lot, and an object could be preferred. Maybe in the future this could be addressed

src/services/sis-redirect-service.ts
```diff
@@ -0,0 +1,52 @@
+import { URL } from "node:url";
+
+type redirectOptions = {
```
umaar: nit: can ignore...but possibly to be consistent with the codebase, you could consider turning this into `RedirectOptions` instead

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (responseFromAuthorizeEndpoint.status === 200) {
+    const authorizationData = await responseFromAuthorizeEndpoint.json();
+    if (!isValidSuccessResponse(authorizationData)) {
+      logger.error("Invalid response properties received from authorization endpoint");
+      throw new Error("Invalid response properties received from authorization endpoint");
+    }
+
+    if (Object.keys(authorizationData.state).length === 0) {
+      logger.error("Response from Authorization API call missing state value");
+      throw new Error("Response from Authorization API call missing state value");
```
umaar: nit: can ignore. But I don't think these can ever fire? Because to have reached line 85, `isValidSuccessResponse` has run, so there will definitely be at least 1 key.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+  const responseFromSessionEndpoint = await fetch(url.href, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+    signal: AbortSignal.timeout(SESSION_TIMEOUT_MS),
+  });
+
+  if (responseFromSessionEndpoint.status === 201) {
```
umaar, line 46: nit: If you look on the `main` branch, we originally had code like this:

```
logger.error(`Session handler returned non-201 status: ${response.status}`);
```

But I think all the status codes have now been lost from the messages which could make debugging tricky. Not sure if that was intentional or done upon request? There are multiple logs like that which have now been lost. We will still get an error message but not a status code. It could mean that a 4xx and a 5xx now look identical.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+    method: "GET",
+    headers: {
+      "session-id": sessionId,
+    },
+  });
+
+  if (responseFromAuthorizeEndpoint.status === 200) {
+    const authorizationData = await responseFromAuthorizeEndpoint.json();
+    if (!isValidSuccessResponse(authorizationData)) {
+      logger.error("Invalid response properties received from authorization endpoint");
```
umaar: nit: Lines 81 and 82 log and throw the same message, and `get-callback-handler.ts` logs the caught error on top, so this ends up in CloudWatch twice per failure. Also same issue on lines 86 and 87 below however there's already an outstanding comment there.

src/services/oauth-internal-service.ts
umaar: nit: Hey @mislam246 not sure if you saw this, let me know if you've got a chance to action that. I'm not sure if Michaela got back to you about what time to reduce it to?

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
umaar: My suggestion was actually to continue using the abort signal consistently and Michaela's suggestion was to potentially remove. Here's the thread https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3612870593

In the end, Michaela agreed to use it consistently, so ideally we keep the testing of an abort signal in the corresponding unit tests, as otherwise that would be untested code

src/services/oauth-internal-service.ts
umaar: nit: Just flagging this, I think this was really asking for a shared private function, but currently we've still got `getRequiredEnvironment("OAUTH_INTERNAL...` in multiple places. One benefit is if it's put into a function, it means that if the variable is ever renamed, we only have to change it in one place.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+export class CreateSessionError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "CreateSessionError";
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+export async function callSessionApi(clientId: string, request: string): Promise<SessionResult> {
+  const oauthInternalApiUrl = getRequiredEnvironment("OAUTH_INTERNAL_API_URL");
```
umaar: nit: Feel free to push back, but I think the spirit of this comment thread: https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3612855081 was possibly not met

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
+import { URL } from "node:url";
+import { getRequiredEnvironment } from "../commons/get-required-environment";
+import logger from "../commons/logger";
+import { isValidSuccessResponse } from "./oauth-internal-service-response";
+
+export type SessionResult = {
```
umaar, line 6: nit: 
Just thinking back to Michaela's point here: https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3612916175

I think it's a good point, but I think the field names haven't changed, so `session_id` and `redirect_uri` still travel to the handlers. The API's naming convention is still leaking through, which I think was the thing we were trying to stop.

Also see the `AuthorizationResult` on line 12: `authorizationCode` is camel case but `redirect_uri` is snake case.

Feel very free to completely ignore this. I don't think it's important right now.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,104 @@
...
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+    },
+    body,
+    signal: AbortSignal.timeout(SESSION_TIMEOUT_MS),
+  });
+
+  if (responseFromSessionEndpoint.status === 201) {
+    const sessionData = await responseFromSessionEndpoint.json();
```
umaar, line 47: nit: We validate the `/api/authorization` response carefully on line 80 but then don't do the same with `/api/session` here. So for example, if there's an unexpected body, `session_id` is `undefined`, I think we could send the user to confirm details with a session that wouldn't work? A type guard could help here, but maybe if things are working, this isn't important

src/handlers/get-callback-handler/get-callback-handler.ts
umaar: nit: Note this does `process.env.DOMAIN_NAME || ""` while the authorize handler uses `getRequiredEnvironment("DOMAIN_NAME")`. Just flagging it now as I guess part of this PR is about consistency, but at the same time, this PR didn't introduce this issue. Also the fallback of an empty string means `sis-redirect-service.ts` can produce `https:///error/unrecoverable` which isn't a valid URL.

Maybe something to improve in the future.

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    },
+    { status: 500 }
+  );
+
+  vitest.stubGlobal("fetch", vitest.fn().mockResolvedValueOnce(mockResponse));
+
+  await expect(
+    getAuthorizationCode("test-client-id", "https://test-uri.com", "test-state", "test-session-id")
+  ).rejects.toThrow("Authorize endpoint returned an error response");
+});
```
umaar, line 194: I'm probs wrong, so treat this as a question for now...but if you look on `main` for `get-callback-handler.test.ts` line 95, there's a 403 test for example which is a good test to have, but then I'm wondering if we lost it in this PR? I haven't cross referenced the different branches in terms of code coverage, but maybe you could possibly double check we haven't _lost_ tests in this PR

src/services/oauth-internal-service.ts
chrisclayson: I agree with @umaar here. Normal pattern for configuration values is to have a module (lets call it `configuration`) that exports functions affecting application configuration values. I would move any and all calls to `process.env.<whatever>` to a shared module and then, as @umaar mentions its easier to update if we ever change the variable name OR switch to using something other than environment variables (for example AppConfig or SSM).

src/services/oauth-internal-service.ts
chrisclayson: The NFR for services is requirement that the P99 response time is < 2.5 seconds. One would hope these calls would not take anywhere near this long, though, the session endpoint does, potentially, need to reach out to JWKS endpoint (if the cache is missed), so I'd be tempted to go for 5 seconds, we want to allow some leeway for random issues and we don't want to break the user journey if the response time exceed the NFR.

When we get to the observability epic, we'll think about adding some monitoring/alerting around the overall `/oauth2/authorize` and the nest `/session` calls so we can identify any potential problems.

I'd also be tempted to pop the timeout in a configuration variable.

src/handlers/get-callback-handler/get-callback-handler.ts
chrisclayson: I'm also wondering if we should favour relative redirects rather than fully qualified ones? 🤔  Then this problem goes away.

Not sure if there are potential (security) issues with that approach though.

src/services/oauth-internal-service.ts
mislam246: I've added a simple function to the `/commons/configuration.ts` file to get the OAuth Internal API URL using the `getRequiredEnvironment()` function that already exists. I've also added a similar function there to retrieve the `SESSION_TIMEOUT_MS` variable.

src/services/oauth-internal-service.ts
mislam246: Change the timeout to 5 seconds and added it to the `/api/authorization` call too. Also moved the timeout into an environment variable.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
mislam246: I've added the signal to both fetches and in the unit tests.

src/services/sis-redirect-service.ts
mislam246: Good point. I've removed the URL decoding from `sis-redirect-service.ts`.

src/services/oauth-internal-service.ts
mislam246: If I've understood correctly, this has now been updated to make sure `OAUTH_INTERNAL_API_URL` is used in both API calls with the new function.

src/services/sis-redirect-service.ts
mislam246: I've removed this ternary operator as suggested. I initially wasn't sure if passing undefined to the `redirect` function would work, but seems like it'll treat it as though it wasn't passed in.

src/services/sis-redirect-service.ts
mislam246: Changed as part of this ticket.

src/services/sis-redirect-service.ts
mislam246: Changed.

src/services/oauth-internal-service.ts
mislam246: Makes sense, I've removed this check and error.

src/services/oauth-internal-service.ts
mislam246: This was unintentionally removed while refactoring. I've added back a status code error log in both the `callSessionApi` and `getAuthorizationCode` functions similar to how it is in main. Those were the only two I think were accidentally removed, unless I've missed others?

src/services/oauth-internal-service.ts
mislam246: True, that causes unnecessary duplication. I've removed the error log from here since it'll be logged in `get-callback-handler.ts` as you mentioned.

src/services/oauth-internal-service.ts
mislam246: I've addressed this in a comment above (https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3690060943), where I've added a function that gets the environment variable.

src/services/oauth-internal-service.ts
mislam246: If there are still some small benefits of decoupling from the APIs, is it worth keeping this as is, instead of undoing the refactor? 

Also, the casing of those variables are how they're returned from the API call, so I don't think it'll work if changed here.

src/services/oauth-internal-service.ts
mislam246: I've added a type guard here to check the response in a similar way to the `/api/authorization` response. Also refactored the `oauth-internal-service-response.ts` to reuse some similar logic, and added unit tests to cover scenarios where the response has missing properties.

src/handlers/get-callback-handler/get-callback-handler.ts
mislam246: Replaced this with `getRequiredEnvironment("DOMAIN_NAME")` and removed the empty string fallback for consistency.

**Edit**: @chrisclayson from what I've read, it seems to be better practice to use fully qualified URLs for redirects. For example, there could be redirect failures where a relative URL isn't resolved correctly (https://http.dev/location).

src/services/tests/oauth-internal-service.test.ts
mislam246: Yeah, it got lost as part of the refactor. I've added it to the `oauth-internal-service.test.ts` given that file handles the status codes now. Also added back the error scenarios for when the authorization endpoint returns missing properties.

src/services/oauth-internal-service.ts
umaar: Defo worth keeping the refactor 👍 

There's was an option for us to change field names, but not a big deal now

src/services/sis-redirect-service.ts
umaar: Nice! I know originally a convo between you and Michaela but if you don't mind me suggesting, I don't think sort is needed, especially as it would also sort the redirect URL which might be a mistake to rearrange? Totally understand if you feel it's not important. I'm just wondering if a redirect URL being changed could ever cause issues, as in theory, we don't own it.

I think the better option is similar to what you're suggesting, move the `state` below the if/else e.g. 

```js
if (authorizationCode) {
  url.searchParams.append("code", authorizationCode);
} else {
  url.searchParams.append("error", "access_denied");
}

url.searchParams.append("state", state);
```

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,290 @@
...
+  );
+
+  expect(jsonSpy).toHaveBeenCalledTimes(1);
+  expect(response.state).toEqual("test-state");
+  expect(response.redirect_uri).toEqual("https://test-uri.com");
+  expect(response.session_id).toEqual("test-session-id");
+});
+
+it("should throw an error if the /api/session fetch returns a 400 status code", async () => {
+  const mockResponse = Response.json({}, { status: 403 });
```
umaar, line 71: nit: notice the title of this test says `400` . Feel free to ignore, not important at all

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
umaar: The authorize test is sorted, thank you. Leaving this open because the same bug is still in `get-callback-handler.test.ts` on line 29, where it's the wrong module name _and_ wrong depth.

src/services/oauth-internal-service.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export class CreateSessionError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "CreateSessionError";
+    Object.setPrototypeOf(this, new.target.prototype);
+  }
+}
+
+const SESSION_TIMEOUT_MS = Number(getSessionTimeout()) || 5000;
```
umaar: A blocker: I think the get callback handler lambda is missing `SESSION_TIMEOUT_MS` and `getRequiredEnvironment` will throw if missing. I think the `SESSION_TIMEOUT_MS` defined in the template should be shared so that both / future lambdas get it automatically. 

Also similarly I guess `|| 5000` isn't doing what we think it does here, as that fallback will never be hit.

I haven't done a deep dive into this but am curious what it take for our tests to properly catch this as I guess everything appears to be passing, but I believe this lambda would totally break! I hope at the very least, our browser tests catch this. I think we should fix this deeper issue (the testing side of things) soon (for example outside of this PR).

src/handlers/get-callback-handler/tests/get-callback-handler.test.ts
```diff
@@ -21,151 +26,61 @@ vitest.mock("@aws-lambda-powertools/logger", () => {
...
-    redirectionURI: "https://api.example.com",
-    authorizationCode: { value: "test-auth-code" },
-    state: { value: "test-state" },
-  });
-
-  globalThis.fetch = vitest.fn().mockResolvedValue({
-    status: 200,
-    json: mockJsonMethod,
-  });
+vitest.mock("../../services/oauth-internal", () => ({
```
umaar: Thanks for addressing this related issue I flagged on the authorise test on the last code review round, but I think it wasn't fixed in this instance (wrong file name being used in this .mock call).

It means the real service file gets used, and that's why `SESSION_TIMEOUT_MS` needed to get set in here whereas `get-authorize-handler.test.ts` didn't need it.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
```diff
@@ -1,20 +1,21 @@
...
 process.env.DOMAIN_NAME = "test-domain";
 process.env.OAUTH_INTERNAL_API_URL = "https://example.com/v1";
 
-const mockFetch = vi.fn();
-vi.stubGlobal("fetch", mockFetch);
-
-afterAll(() => {
-  vi.unstubAllGlobals();
-});
+vi.mock("../../../services/oauth-internal-service", () => ({
```
umaar, line 9: Please do treat this as a small observation: I might be getting confused. However if this observation is correct, it might be the case that this test is passing for the wrong reason.

This line replaces the service with a fake, but the fake only contains `callSessionApi`. But then line 5 imports `CreateSessionError` which the fake doesn't have. So later, on line 111 when we do `throw new CreateSessionError("Session endpoint returned an error response");` that class doesn't exist and vitest will actually throw, then the handler catches it and the test passes. So I think the test is passing accidentally / giving us a false sense of security

src/services/oauth-internal-service.ts
umaar: Thank you, nice.

I think this has bought on a blocking issue: https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3710999016 so maybe worth leaving this thread open as an extra reminder

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,290 @@
...
+  vitest.stubEnv("OAUTH_INTERNAL_API_URL", "https://test.com");
+});
+
+afterEach(() => {
+  vitest.restoreAllMocks();
+  vitest.unstubAllEnvs();
+  vitest.unstubAllGlobals();
+});
+
+it("should call the /api/session fetch and return the raw response", async () => {
```
umaar: A small nit, feel free to ignore: but where it says `...and return the raw response"` is that defo the case I'm wondering. Same on line 80, I think they now return SessionResult / AuthorizationResult

src/services/tests/oauth-internal-service.test.ts
```diff
@@ -0,0 +1,290 @@
+import { afterEach, beforeEach, expect, it, vitest } from "vitest";
```
umaar, line 1: Overall really nice work on the refactoring. One extra thing I noticed in this whole test file, fetch always succeeds, meaning we're not testing when fetch fails, which it can do for many reasons - one being the timeout stuff we added in.

We _did_ have tests covering this in `main` so maybe in a sense we have less test coverage now: https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/main/src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts#L166

`get-callback-handler.test.ts` also tested a failing fetch in `main` (see line 158 on main)

src/services/oauth-internal-service.ts
mislam246: I've addressed your other comment on the blocking issue and left a question.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
mislam246: Fixed this in your other comment (https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/396#discussion_r3711182570).

src/services/sis-redirect-service.ts
mislam246: Fair point. I've removed the `.sort()` and moved the `state` appending after the if/else.

src/services/tests/oauth-internal-service.test.ts
mislam246: Changed the description to `403` instead of `400`.

src/services/oauth-internal-service.ts
mislam246: Added the environment variable to the `GetCallbackHandler` resource in the `template.yaml` file. To make it more reusable, I created a Parameter for the timeout value and referenced that in both resources. I didn't add it to the Globals to prevent all other resources picking up the variable.

Regarding the default fallback, I think you're correct that it wouldn't be hit since an error would be thrown if the variable isn't set. I've removed the default.

With regards to the scenario where the variable isn't set, I assume we want to redirect to the SIS error page. If that's the case, it may be better to call `getSessionTimeout()` from within `callSessionApi()` and `getAuthorizationCode()` instead of how it's called at the module level currently. This way, if any errors are thrown, it'll automatically redirect to the SIS error page. Also a similar test can be added to check it redirects. What are your thoughts @umaar and @chrisclayson?

src/handlers/get-callback-handler/tests/get-callback-handler.test.ts
mislam246: Fixed, thanks for pointing it out.

src/handlers/get-authorize-handler/tests/get-authorize-handler.test.ts
mislam246: Good observation. Since the handler processes all errors the same way, the test is working, which is misleading as you mentioned. I've removed the import and thrown a generic error in the test instead of using the CreateSessionError now.

src/services/tests/oauth-internal-service.test.ts
mislam246: Good point, forgot to update the descriptions. I've amended them both now.

src/services/tests/oauth-internal-service.test.ts
mislam246: Thanks. That test still exists in `get-authorize-handler.test.ts`, line 134, where it mocks the fetch failing. it was lost from the `get-callback-handler.test.ts` which I've added back in. Both handlers test this scenario, so I don't think it's worth adding a similar test to the `oauth-internal-service.ts`. Happy to add it though if you feel it's needed.

src/services/oauth-internal-service.ts
chrisclayson: I think you need to think about the Lambda lifecycle here.

Doing it at the module level, as-is, will cause the initialisation of the lambda to fail and the runtime fail to start. This will trigger the lambda error alarm which, in turn, will trigger a rollback of the misconfigured code.

Doing it the in the handler, as you suggest, will cause user facing errors and the user to be redirected to the error page.

As these values are being pulled from environment variables, which can't change once deployed, this would indicate a misconfiguration and the automatic rollback would be preferred.

If we were passing these values in using AppConfig via the configuration repository then, as configuration change outside the lambda runtime, I would agree catching it and redirecting to the error page would be preferable.

chrisclayson: In short, its fine as it is 👍🏻 

If we move the config into AppConfig, then we should review the position.

chrisclayson (approved):

umaar (approved):

PR #408, opened 2026-07-22
SPT-1994: Add Authorizer function

mislam246:
## What

- Create API Gateway Authorizer function to validate the token passed in the request to the user-identity handler
- Add a Lambda to query the session table using the GSI that returns the `subject` and `storageToken` to be passed in the context to the user-identity handler
-  Add feature to test the unhappy paths (the happy path is tested in end to end browser tests).

Note, the rationale for the logic in the function is to return an `Allow` policy for valid `Bearer` tokens but to throw an `Unauthorized` exception for all other cases. The `Unauthorized` exception, by default, will translate to a `401` error code. We will not return `Deny` policies, at this stage, as this results in a `403` (i.e. "we know who you are, but you don't have permissions").

All other (unexpected) exceptions from the function will result in a `5xx` and will be graphed and alarmed upon, as they will need to be investigated.

## Why

To ensure token passed in the request is valid before redirecting the request to the get-user-identity-handler.


infrastructure/identity-reuse-service/template.yaml
```diff
@@ -1521,6 +1523,18 @@ Resources:
...
+  PrivateApiAuthorizerResourcePolicy:
+    Type: AWS::Lambda::Permission
+    Condition: IsNotProduction
+    Properties:
+      FunctionName:
+        Fn::ImportValue: !Sub "${OauthInternalStackName}-ProtectedResourceAuthorizerFunctionArn"
+      Action: lambda:InvokeFunction
+      Principal: apigateway.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
+      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PrivateApi}/authorizers/*"
```
chris-dco, line 1536: Does this need a `*` at the end?

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -1609,6 +1628,19 @@ Resources:
...
+    Condition: IsDev
+    Properties:
+      FunctionName:
+        Fn::ImportValue: !Sub "${OauthInternalStackName}-ProtectedResourceAuthorizerFunctionArn"
+      Action: lambda:InvokeFunction
+      Principal: apigateway.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
+      # The presence of * in the sourceARN is, sadly, required as there seems to be no way to get
+      # the ARN or ID of the authorizer resource generated by the OpenAPI spec
+      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PrivateApiDev}/authorizers/*"
```
chris-dco, line 1640: As [above](https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/408/changes#r3674372893) is the wildcard needed?

src/commons/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class PolicyGenerationError extends Error {
+  constructor(message: string | undefined) {
```
chris-dco: Just a suggestion, but you can make this more concise.

```suggestion
  constructor(message?: string) {
```

src/commons/hashing.ts
```diff
@@ -0,0 +1,7 @@
+import { createHash } from "node:crypto";
+
+export const sha256Hash = (value: string): string => {
+  const hash = createHash("sha256");
```
chris-dco: As a suggestion the `createHash` and resulting Hash always returns the Hash object. So it is quite common to see:

```suggestion
  return createHash("sha256").update(value).digest("hex");
```

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+export const handler = async (
+  event: APIGatewayRequestAuthorizerEvent,
+  context: Context
+): Promise<APIGatewayAuthorizerResult> => {
+  logger.addContext(context);
+  const headers: APIGatewayRequestAuthorizerEventHeaders = event.headers ? lowerCaseHeaderKeys(event.headers) : {};
+
+  if (!headers["authorization"]) {
+    logger.error("Authorization failed as no header present");
+    addMetric("AccessTokenValidationFailure", 1, {
```
chris-dco: Can you add the metric to `src/commons/metric-enum.ts`?

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+      reason: "header-not-present",
+    });
+    throw new Error("Unauthorized");
+  }
+
+  if (!headers["authorization"]?.startsWith("Bearer ")) {
+    logger.error("Authorization header is not a bearer token");
+    addMetric("AccessTokenValidationFailure", 1, {
```
chris-dco: Can you reference metrics from `src/commons/metric-enum.ts`?

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+): Promise<APIGatewayAuthorizerResult> => {
+  logger.addContext(context);
+  const headers: APIGatewayRequestAuthorizerEventHeaders = event.headers ? lowerCaseHeaderKeys(event.headers) : {};
+
+  if (!headers["authorization"]) {
+    logger.error("Authorization failed as no header present");
+    addMetric("AccessTokenValidationFailure", 1, {
+      reason: "header-not-present",
+    });
+    throw new Error("Unauthorized");
```
chris-dco, line 38: Normally if the `user-identity` function detects an error it's redirected to the unrecoverable error page. In the case of the authorizer rejecting the request what will happen to the user? Will they be redirected to the error page or get a vanilla AWS 500 error code?

Shouldn't the authorizer return a valid APIGatewayAuthorizerResult such as the following:
```json
{
  "principalId": "user",
  "policyDocument": {
    "Version": "2012-10-17",
    "Statement": [
      {
        "Action": "execute-api:Invoke",
        "Effect": "Deny",
        "Resource": "arn:aws:execute-api:*:*:*"
      }
    ]
  }
}
```

I then understand that the OpenAPI specification can then contain the following code which would cause the redirect:
```yaml
x-amazon-apigateway-gateway-responses:
  ACCESS_DENIED:
    statusCode: 302
    responseParameters:
      gatewayresponse.header.Location: "'https://yourdomain.com/error/unrecoverable'"
    responseTemplates:
      application/json: ""
```

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  if (!headers["authorization"]?.startsWith("Bearer ")) {
+    logger.error("Authorization header is not a bearer token");
+    addMetric("AccessTokenValidationFailure", 1, {
+      reason: "not-a-bearer-token",
+    });
+    throw new Error("Unauthorized");
```
chris-dco, line 46: Ref https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/408/changes#r3674500411

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+      });
+    }
+  } catch (error) {
+    logger.error("Unexpected error in API Gateway handler", { error });
+    addMetric("AccessTokenValidationFailure", 1, {
+      reason: "exception",
+    });
+    throw new Error("Unexpected error", { cause: error });
+  }
+  throw new Error("Unauthorized");
```
chris-dco, line 94: Ref https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/408/changes#r3674500411

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+};
+
+export const lowerCaseHeaderKeys = function (
```
chris-dco, line 97: The `src/handlers/post-phase2-user-identity-handler/post-phase2-user-identity-handler.ts` has generic could which could be migrated to src/commons to do this and would make it reusable across lambdas.

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,132 @@
...
+    const policyDocument: PolicyDocument = { Version: "2012-10-17", Statement: [statement] };
+    logger.debug("Generating policy " + effect);
+    const context = {
+      subjectId,
+      storageToken,
+    };
+    return { principalId, policyDocument, context };
+  }
+
+  throw new PolicyGenerationError("Missing effect and/or resource to generate policy.");
```
chris-dco, line 133: Is there a case when this would actually happen?

tests/acceptance/features/useridentity-GET-UnhappyPath.feature
```diff
@@ -0,0 +1,16 @@
+Feature: UserIdentity Get - Happy Path
```
chris-dco: As with https://github.com/govuk-one-login/ipv-identity-reuse-service/pull/408/changes#r3674500411 we need to consider the user journey. As /user-identity is a page than can be navigated to by the user (rather than a service) the user should be redirected to a page allowing them to recover rather than an unfriendly page.

infrastructure/identity-reuse-service/template.yaml
chrisclayson: Sadly, yes 😢 

I intended to make this better, but the core problem is to make specific to the API Gateway authoriser, you need the authoriser ID (instead of the `*`), but in an API generated by an OpenAPI spec you can neither:
- Get a reference to the authorizers auto-generated by the spec
- Separately define the authorizer and reference within OpenAPI

It may be that, in this case, we resort to giving API Gateway an execution role, rather than relying on the Lambda resource policy. For now, I think we leave as is, but look to improve before any security testing.

tests/acceptance/features/useridentity-GET-UnhappyPath.feature
chrisclayson: `/user-identity` isn't a page, its an API call. The user is orchestration.

As such, no redirects required. It's 401 or 403.

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
chrisclayson: Potentially if, for some reason, the resource (i.e. `context.methodArn`) wasn't passed in correctly 🤷🏻

infrastructure/identity-reuse-service/template.yaml
chrisclayson: I did put the comment above it ... though, maybe I take that out.

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
chrisclayson: I also considered ripping this out, as API Gateway lowercases everything on the way through anyway.

chris-dco (dismissed):

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,134 @@
...
+  event: APIGatewayRequestAuthorizerEvent,
+  context: Context
+): Promise<APIGatewayAuthorizerResult> => {
+  logger.addContext(context);
+  const authorizationHeader = getProperty(event.headers || {}, "authorization");
+
+  if (!authorizationHeader) {
+    logger.error("Authorization failed as no header present");
+    addMetric(MetricName.AccessTokenValidationFailure, 1, {
+      Reason: "header-not-present",
```
chris-dco: This should also be:

```suggestion
      [MetricDimension.Reason]: "header-not-present",
```

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,134 @@
...
+    });
+    throw new Error("Unauthorized");
+  }
+
+  if (!authorizationHeader?.startsWith("Bearer ")) {
+    logger.error("Authorization header is not a bearer token");
+    addMetric(MetricName.AccessTokenValidationFailure, 1, {
+      Reason: "not-a-bearer-token",
```
chris-dco:
```suggestion
      [MetricDimension.Reason]: "not-a-bearer-token",
```

src/handlers/api-gateway-protected-resource-authorizer-handler/api-gateway-protected-resource-authorizer-handler.ts
```diff
@@ -0,0 +1,134 @@
...
+  eventHeaders: APIGatewayRequestAuthorizerEventHeaders
+): APIGatewayRequestAuthorizerEventHeaders {
+  const headers: APIGatewayRequestAuthorizerEventHeaders = {};
+  for (const key of Object.keys(eventHeaders)) {
+    headers[key.toLowerCase()] = eventHeaders[key];
+  }
+  return headers;
+};
+
+function addMetric(metricName: string, value: number = 1, dimensions?: Dimensions) {
```
chris-dco:
```suggestion
function addMetric(metricName: MetricName, value: number = 1, dimensions?: Dimensions) {
```

chrisclayson: Not convinced this actually works.

chrisclayson: `MetricName` isn't actually an `enum` its an object of strings.

chrisclayson: Looks like converting to an `enum` works 😄

chris-dco (approved):

PR #411, opened 2026-07-23
SPT-2020: Add Lambda permissions for API Gateway

umaar:
## What

- SPT-2020: Add Lambda invoke permissions for API Gateway
- Fixes a previously incorrect `SourceArn`

## Why

- In SIS, we seem to have a mixture of ways of giving API Gateway the permission to execute a lambda. This begins to use the correct approach. Follow up PR is incoming.

### DIFF
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -538,6 +538,16 @@ Resources:
         - Key: Name
           Value: !Sub "/aws/lambda/${AWS::StackName}-GetAuthorizeHandler"
 
+  PublicApiGetAuthorizeHandlerResourcePolicy:
+    Condition: IsNotProduction
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref GetAuthorizeHandler.Alias
+      Action: lambda:InvokeFunction
+      Principal: apigateway.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
+      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PublicApi}/${ApiStageName}/GET/oauth2/authorize"
+
   ################################
   # GET USER IDENTITY HANDLER    #
   ################################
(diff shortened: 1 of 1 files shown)


infrastructure/identity-reuse-service/template.yaml
```diff
@@ -1791,25 +1868,17 @@ Resources:
...
-        - ":"
-        - - "function"
-          - !Select [6, !Split [":", !GetAtt PostPhase2UserIdentityHandler.Arn]]
-          - "live"
+      FunctionName: !Ref PostPhase2UserIdentityHandler.Alias
       Action: lambda:InvokeFunction
       Principal: apigateway.amazonaws.com
       SourceAccount: !Ref AWS::AccountId
-      SourceArn: !Sub "arn:aws:apigateway:${AWS::Region}::/restapis/${DevPrivateRestApi}/${ApiStageName}"
+      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${DevPrivateRestApi}/${ApiStageName}/POST/*"
```
umaar: I don't believe the old version would've even worked. This fixes it.

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -538,6 +538,16 @@ Resources:
...
 
+  PublicApiGetAuthorizeHandlerResourcePolicy:
+    Condition: IsNotProduction
+    Type: AWS::Lambda::Permission
+    Properties:
+      FunctionName: !Ref GetAuthorizeHandler.Alias
+      Action: lambda:InvokeFunction
+      Principal: apigateway.amazonaws.com
+      SourceAccount: !Ref AWS::AccountId
+      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PublicApi}/${ApiStageName}/GET/*"
```
chrisclayson:
```suggestion
      SourceArn: !Sub "arn:aws:execute-api:${AWS::Region}:${AWS::AccountId}:${PublicApi}/${ApiStageName}/GET/oauth2/authorize"
```

To make these _uber_ secure, we should limit to just the endpoint that should be executing the lambda, in this case replace the `/*` with `/oauth2/authorize`.

Same applies to the other permissions.

infrastructure/identity-reuse-service/template.yaml
chrisclayson: Yeah, good catch, I don't think it would have worked either 🙈

infrastructure/identity-reuse-service/template.yaml
umaar: Oh nice! Done

chrisclayson (approved):
LGTM

chris-dco (approved):

PR #412, opened 2026-07-24
SPT-1968: Updated Post confirm details lambda to redirect to the GET /oauth2/common endpoint

chris-dco:
## What

Updated the following endpoints to enable the Post conform details lambda to redirect to the `GET /oauth2/common` endpoint:
* `get-authorize-handler` and get-confirm-details-handler now passes through the client_id downstream.
* `post-confirm-details-handler` now calls the /oauth2/common endpoint with the client_id.

Updated CloudFormation to include the following parameters for the `oauth-internal` stack.
* `OrchestrationJwksEndpointOverride` - Override for the default `OrchestrationJwksEndpoint` for the environment.
* `OrchestrationRedirectURIOverride` - Override for the default `OrchestrationRedirectURI` for the environment.

## Why

* Lambda updates - Following the Confirm Details screen the user should be redirected to the `/oauth2/common` endpoint to continue their journey. The `/oauth2/common` endpoint requires the `client_id` parameter which was not being passed through the different steps.

* Cloud Formation updates - The Cloud Formation stacks contained hard coded values to the Stored Identity Services Public and Private APIs. This made it very hard from a developer perspective to deploy their own copy of the Stored Identity Service as they would need to update the Cloud Formation directly. Whereas the new parameters enable a very minor change on the command line or a developers `samconfig.toml` file.

### Steps to Test

The following was done to test this functionality:
1. Deployed Stored Identity Service to environment
1. Deployed Orchestration Stub using the outputs from the Stored Identity Service
1. Ran a journey using the defaults
1. At the "You have already proved your identity" screen:
  1. Opened the browser development tools
  2. Look for the value of the `identity_reuse_service_session`
  3. Look at OAuth Common DynamoDb session table and where the `sessionId` is the value of `identity_reuse_service_session`
  4. Edited the record adding an `authorizationCode` with any value
1. Clicked Continue

#### Image showing DynamoDb record:

Screenshot 2026-07-24 at 08 35 59

#### Image showing final screen include `code` in the address bar:

Screenshot 2026-07-24 at 08 43 08

### DIFF
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -84,7 +84,7 @@ Parameters:
   OauthInternalStackName:
     Description: OAuth Internal Stack name
     Type: String
-    Default: "sis-oauth-internal"
+    Default: preview-main-oauth
 
   DynatraceNodeLayerArn:
     Type: String
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -84,7 +84,7 @@ Parameters:
   OauthInternalStackName:
     Description: OAuth Internal Stack name
     Type: String
-    Default: preview-main-oauth
+    Default: "sis-oauth-internal"
 
   DynatraceNodeLayerArn:
     Type: String
@@ -1075,6 +1075,7 @@ Resources:
       Environment:
         Variables:
           NODE_OPTIONS: "--enable-source-maps"
+          PUBLIC_API: !GetAtt ReuseIdentityDomain.Value
     Metadata:
       BuildMethod: esbuild
       BuildProperties:
--- a/src/handlers/post-confirm-details-handler/post-confirm-details-handler.ts
+++ b/src/handlers/post-confirm-details-handler/post-confirm-details-handler.ts
@@ -11,9 +11,14 @@ export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGat
     throw new Error("One or more required query string parameters are undefined");
   }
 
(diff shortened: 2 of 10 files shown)


infrastructure/identity-reuse-service/template.yaml
```diff
@@ -84,7 +84,7 @@ Parameters:
   OauthInternalStackName:
     Description: OAuth Internal Stack name
     Type: String
-    Default: "sis-oauth-internal"
+    Default: preview-main-oauth
```
michaelabenyohai: My understanding from @chrisclayson was that this had to stay as `sis-oauth-internal` because that is what it is named in Build upwards. And the pipelines cannot override this value, whereas in dev we can.

Though now that the stacks already exist in Build upwards maybe this isn't a problem. What do you think?

src/handlers/post-confirm-details-handler/post-confirm-details-handler.ts
```diff
@@ -11,9 +11,14 @@ export const lambdaHandler = async (event: APIGatewayProxyEvent): Promise<APIGat
     throw new Error("One or more required query string parameters are undefined");
   }
 
-  const url = new URL(decodeURIComponent(eventValues.get("redirectUri") || ""));
-  url.searchParams.append("code", "SplxlOBeZQQYbYS6WxSbIA");
-  url.searchParams.append("state", eventValues.get("state") || "");
+  // const url = new URL(decodeURIComponent(eventValues.get("redirectUri") || ""));
+  // url.searchParams.append("code", "SplxlOBeZQQYbYS6WxSbIA");
+  // url.searchParams.append("state", eventValues.get("state") || "");
```
michaelabenyohai: Is it worth deleting this commented out code?

michaelabenyohai (approved):

PR #416, opened 2026-07-29
SPT-2032: Adds browser test skeleton

umaar:
## What

- Adds browser test skeleton with playwright
- Also adds GitHub actions workflow to run them
- Adds `eslint-plugin-playwright` without failing build

## Why

- The acceptance tests should test, end-to-end, the various flows through SIS. They should test from a user perspective, in the case of SIS the user will be using a browser, so using browser driven tests allows us to prove the journeys work.


eslint.config.mjs
```diff
@@ -1 +1,23 @@
-export { default } from "@govuk-one-login/ipv-trust-and-reuse-eslint-rules";
+import { defineConfig } from "eslint/config";
+import playwright from "eslint-plugin-playwright";
+import sharedRules from "@govuk-one-login/ipv-trust-and-reuse-eslint-rules";
+
+const playwrightRecommended = playwright.configs["flat/recommended"];
+
+// Advises without failing the build.
+const advisoryRules = Object.fromEntries(
+  Object.entries(playwrightRecommended.rules ?? {}).map(([rule, level]) => [rule, level === "off" ? "off" : "warn"])
```
umaar, line 9: We can ditch `eslint-plugin-playwright`, but the hope/intention was that the rules in that package will helpful in maintaining our browser tests to a high standard., and thought this was a decent middle ground - so it's included but won't fail the build. 

I haven't (yet?) gone through every rule part of `eslint-plugin-playwright` - it might be we don't want any/some of those rules. But at a quick glance, some of the rules there look very sensible.

.github/workflows/browser-tests.yml
```diff
@@ -0,0 +1,71 @@
+name: browser-tests
```
chris-dco: I think this PR is great. However, I think we should remove the changes to the .github/workflows for now. Sorry for the following long explanation which I think we need to spin out into another ticket.

Currently the Orchestration Stub, Stored Identity Store and OAuth Common are configured with a circular dependency. Where Orchestration Stub calls the Private/Public APIs of the Stored Identity Service, Stored Identity Service calls OAuth Common and OAuth Common reads its jwks.json file from the Orchestration Stub. This means that these three components are tightly coupled to one another.

The net result is that if your tests call https://orch.reuse.dev.stubs.account.gov.uk/ they will always call the main deployments of Orchestration Stub, Stored Identity Service and OAuth common. This has been one of the main things which has given me a headache over the last 3 weeks as you need to know how to deploy each of these in order to develop just one.

This makes testing these components incredibly challenging.

tests/acceptance/browser/authorization-workflow.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { expect, test } from "./fixtures";
+import { ConfirmDetailsPage } from "./pages/confirm-details.page";
+import type { AuthorizationRequestField } from "./pages/orchestration-stub.page";
+import { UnrecoverableErrorPage } from "./pages/unrecoverable-error.page";
+
+const AN_HOUR = 60 * 60;
```
chris-dco, line 6: Do you think it would make it clearer to add the unit of measurement `_SECONDS` after `AN_HOUR` and `TEN_MINUTES`.

eslint.config.mjs
chris-dco: May be worth adding it to @govuk-one-login/ipv-trust-and-reuse-eslint-rules in the future. When we're more familiar with it.

chris-dco (commented):
Looks very good and I look forward to using it. The only except for one small suggestion I think my only concern is relating to running the tests in GitHub actions. Also, would it be possible to spin off another ticket to run these as part of the acceptance tests in build?

.github/workflows/browser-tests.yml
umaar: Totally fair point! Shall I remove it now, or did you want other people to chime into that team discussion?

tests/acceptance/browser/authorization-workflow.spec.ts
umaar: As in `AN_HOUR_SECONDS` ? Probs just me but that would confuse me! Whatever works though, just sharing my thoughts, but if you think it'd be better, more than happy to change it

umaar:
Thanks for the feedback!

- For GitHub actions: Happy to do wait for that team discussion you raised to reach a conclusion. Edit: Removed
- Have raised that ticket you mentioned: SPT-2038

chris-dco (approved):

PR #417, opened 2026-07-29
SPT-1968: PostConfirmDetails call CreateAuthCodeFn

chris-dco:
## What

Updated the PostConfirmDetails Lambda to call the CreateAuthConfFn Lambda in OAuth common. This involved the following changes:
* Added missing `POWERTOOLS_SERVICE_NAME` and `POWERTOOLS_METRICS_NAMESPACE` to some lambdas which was causing CloudWatch to log undefined for the service name.
* Updated OAuth Common SAR version to v0.8.0 as this provides to new `CreateAuthCode` function which is required by `PostConfirmDetailsHandler`
* Added the `/api/create-auth-code` API to the OAuth Common API Gateway and wired this up to the `CreateAuthCodeFn` lambda exported by OAuth Common SAR version v0.8.0.
* Updated the `PostConfirmDetailsHandler` to:
  * Be able to access the OAuth Internal Stack
  * Call the `/api/create-auth-code` API to generate an auth code
  * Redirect to the error page in the event of an error
* Updated the `GetCallbackHandler`:
  * Cleaned up the handler to make it follow the same pattern as other lambdas for the `OAUTH_INTERNAL_API_URL` environment variable.
  * Moved the code to redirect into the `try ... catch`. If the lambdas is incorrectly configured an unhandled exception is thrown if the `OAUTH_INTERNAL_API` is not defined.
* Moved `redirectToErrorPage` and `redirect` functions to common/redirect-utilities.ts so that it can be reused by different lambdas.
* Added various bit of error handling and metrics generation to make debugging easier.

## Why

In order to enable SIS to make use of Common OAuth components, SIS and it’s Orchestration stub will require updates to integrate into the OAuth flow. This ticket ensures that an authentication code is generated and stored in the session table against the session ID after the user clicks “continue” on `/confirm-details`.

## Testing

1. Navigate to https://orch-spt-1968.reuse.dev.stubs.account.gov.uk which is been deployed to test this functionality.
2. At the bottom of the page click Continue
3. The page "Orchestration stub: Response" should appear and the URL will contain the authorization code generated in the Session Table

### DIFF
--- a/infrastructure/identity-reuse-service/template.yaml
+++ b/infrastructure/identity-reuse-service/template.yaml
@@ -284,6 +284,7 @@ Globals:
       - arm64
     Environment:
       Variables:
+        POWERTOOLS_METRICS_NAMESPACE: !GetAtt MainCloudWatchMetricNamespaceName.Value
         AWS_LAMBDA_EXEC_WRAPPER: !If [DynaTraceEnabled, "/opt/dynatrace", !Ref "AWS::NoValue"]
         DT_CONNECTION_AUTH_TOKEN: !If
           - DynaTraceEnabled
@@ -375,8 +376,6 @@ Resources:
       Environment:
         Variables:
           METRIC_SERVICE_NAME: "GetAuthorizeHandler"
-          POWERTOOLS_METRICS_NAMESPACE:
-            !If [IsDev, !Sub "${AWS::StackName}-GetAuthorizeHandler", "GetAuthorizeHandler"]
           POWERTOOLS_SERVICE_NAME: "GetAuthorizeHandler"
           ENVIRONMENT: !If [IsDev, "dev", !Ref Environment]
           APP_CONFIG_NAME: !Ref ApplicationConfigName
@@ -563,8 +562,6 @@ Resources:
           METRIC_SERVICE_NAME: "GetUserIdentityHandler"
           EVCS_API_KEY_SECRET_ARN:
             Fn::ImportValue: !Sub "${SharedStackName}-EVCSApiKeySecret"
-          POWERTOOLS_METRICS_NAMESPACE:
-            !If [IsDev, !Sub "${AWS::StackName}-GetUserIdentityHandler", "GetUserIdentityHandler"]
           POWERTOOLS_SERVICE_NAME: "GetUserIdentityHandler"
           ENVIRONMENT: !If [IsDev, "dev", !Ref Environment]
           APP_CONFIG_NAME: !Ref ApplicationConfigName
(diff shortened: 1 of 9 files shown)


infrastructure/identity-reuse-service/template.yaml
```diff
@@ -954,6 +954,9 @@ Resources:
       Environment:
         Variables:
           NODE_OPTIONS: "--enable-source-maps"
+          POWERTOOLS_SERVICE_NAME: "GetConfirmDetailsHandler"
+          POWERTOOLS_METRICS_NAMESPACE:
```
chrisclayson: I do think that the namespace for all our metrics should be the same. 

Personally, I'd be happy with `POWERTOOLS_METRICS_NAMESPACE: !Ref AWS::StackName` see what I did [here](https://github.com/govuk-one-login/ipv-identity-reuse-service/blob/5cefa746f1b17dd0d495f5cc5eb5987a62194ec3/infrastructure/oauth-internal/template.yaml#L504) with the authoriser.

chrisclayson: In our other products, metrics are all in the same namespace, for example: `POWERTOOLS_METRICS_NAMESPACE: !If [ IsDev, !Sub "${AWS::StackName}-<product-name>", "<product-name>" ]`

chris-dco: I have added the following to the globals so that it's inheritted for each lambda:
```yaml
Globals:
  Function:
    ...
    Environment:
      Variables:
        POWERTOOLS_METRICS_NAMESPACE: !GetAtt MainCloudWatchMetricNamespaceName.Value
```

And also added the following SSM parameter which enables reuse between different resource without needing to redefine it:

```yaml
  MainCloudWatchMetricNamespaceName:
    Type: AWS::SSM::Parameter
    Properties:
      Type: String
      Value: !If
        - IsDev
        - !Sub "${AWS::StackName}-${MainCloudWatchMetricNamespace}"
        - !Ref MainCloudWatchMetricNamespace
```

infrastructure/identity-reuse-service/template.yaml
```diff
@@ -284,6 +284,7 @@ Globals:
       - arm64
     Environment:
       Variables:
+        POWERTOOLS_METRICS_NAMESPACE: !GetAtt MainCloudWatchMetricNamespaceName.Value
```
chrisclayson, line 287: Happy to approve, as is, but just wondering why opting to use an SSM parameter here rather than just  do the calculation here?

chrisclayson (approved):

infrastructure/identity-reuse-service/template.yaml
chris-dco: Reduce duplication. I've seen too many Cloud Formation files where calculations are duplicated over and over and often one of them is wrong.

PR #422, opened 2026-07-30
SPT-1994: Do not alert on lambda auth errors in build

chrisclayson:
## What

Disable the alarm actions for the lambda authoriser to prevent alarms being raised in dev and build.

## Why

The alarms are going off for "Unauthorized" errors, which we don't want. We may need to review how the authoriser works and interacts with the alerting, but for the immediate term, lets disable the alerting.

## Related PRs

#408


umaar (dismissed):

umaar:
Approved, but note the linter `  [[W8001: Check if Conditions are Used] (Condition IsNotDev not used) matched 177]`

umaar (approved):

PR #423, opened 2026-07-31
SPT-1968: Deploy script deploys orch stub

chris-dco:
## What

Added the ability for the deploy script to also deploy the orchestration stub.

## Why

There is a cyclic dependency between Orchestration, SIS and OAuth common so we need to be able to deploy all three of these and feed in our outputs from each stack into the other.

### DIFF
--- a/deploy-to-dev.sh
+++ b/deploy-to-dev.sh
@@ -13,6 +13,8 @@ Usage:
     -s      --stack-name        The name of the stack to deploy.
     -e      --environment       Environment that the stack is being deployed to. (default = 'local')
     -o      --oauth-deploy      Deploy the oauth-internal stack (optional)
+    -c      --orch-deploy       Deploy the orchestration-stub stack (optional)
+    -p      --stubs-profile     The AWS profile for the stubs environment. (default = 'stubs-dev')
     -n      --no-sis-deploy     Do not deploy the identity-reuse-service stack (optional)
     -m      --shared-stack-name The name of an existing identity-reuse-shared stack that the oauth-internal can point
                                 at for audit infrastructure. (optional)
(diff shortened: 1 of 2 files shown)


deploy-to-dev.sh
```diff
@@ -27,170 +29,296 @@ EOF
...
+  AWS_DEFAULT_REGION=eu-west-2
+  SAM_CMD=sam
 
-  SAM_CONFIG="infrastructure/${1}/samconfig.toml"
-  [[ -e "${SAM_CONFIG}" ]] || RESOLVE_S3=true
-  if "${RESOLVE_S3}"; then
-    BUCKET_PARAM=("--resolve-s3")
-  else
-    BUCKET_PARAM=("--config-file" "../../../$SAM_CONFIG")
+  if [[ "${OSTYPE}" =~ ^msys ]]; then
```
joe-mcgarry, line 47: How come we need this check adding in? Are some folks using an msys operating system?

deploy-to-dev.sh
```diff
@@ -27,170 +29,296 @@ EOF
...
+
+  pushd "${BASE_DIR}"
+  resolve_bucket_param "${component}"
 
-export AWS_DEFAULT_REGION=eu-west-2
-echo "Environment: ${ENVIRONMENT}"
-echo "Profile:     ${AWS_PROFILE}"
-$DEPLOY_SIS && echo "Deploy identity-reuse-service to stack ${SIS_STACK_NAME}, pointing at shared stack ${SHARED_STACK_NAME}"
-$DEPLOY_OAUTH && echo "Deploy oauth-internal to stack ${OAUTH_STACK_NAME}, pointing at shared stack ${SHARED_STACK_NAME}"
+  echo -e "\n\033[1;34m==> Deploying component '${component}' to stack '${stack_name}'\033[0m\n"
```
joe-mcgarry, line 195: Is this you showing off your ASCII formatting skills? 😅

chrisclayson (approved):

deploy-to-dev.sh
chris-dco: I may have been learning to do it. With some help from Kiro as I wanted it clearer to see what was going on.

PR #432, opened 2026-08-11
SPT-2025: Add cryptographic checks to confirm details handler

joe-mcgarry:
## What

- Refactors the cryptographic validation logic from `post-phase2-user-identity-handler`, extracts it into `src/commons/validate-records.ts`
- Uses extracted logic in `get-confirm-details-handler`
- Adds `TokenValidationError` and `CredentialStoreError` so each handler can deal with error paths slightly more elegantly 
- The various in-line validations done to the `kid`, `signature` and checking the validity are now done in `validateIdentityRecords` in `validate-records` in a single call

## Why

Logic for validating the VC signatures in both handlers would be duplicated, so the logic was extracted and code refactored to minimise duplication

### DIFF
--- a/shared-test/credential-store-utilities.ts
+++ b/shared-test/credential-store-utilities.ts
@@ -1,0 +1,47 @@
+import { JWTHeaderParameters } from "jose";
+import { CredentialStoreIdentityResponse } from "../src/credential-store/credential-store-identity-response";
+import { getDefaultJwtHeader, sign } from "./jwt-utilities";
+import { Mock } from "vitest";
+
+const CURRENT = "CURRENT";
+
+export const createStoredIdentityRecord = (...credentialSignatures: string[]) => {
+  const base = { sub: "user-sub", vot: "P2", vtm: "https://oidc.account.gov.uk/trustmark" };
+  return credentialSignatures.length > 0 ? { ...base, credentials: credentialSignatures } : base;
+};
+
+export const createSignedIdentityCheckCredentialJWT = async (issuer: string, nbfDate?: string): Promise<string> => {
+  const nbf = Math.floor((nbfDate ? new Date(nbfDate).getTime() : Date.now()) / 1000);
+  return await sign(getDefaultJwtHeader(), {
+    iss: issuer,
+    nbf,
+    sub: "sdf",
+    vc: { evidence: [{}], type: ["VerifiableCredential", "IdentityCheckCredential"] },
+  });
+};
+
+export const createCredentialStoreIdentityResponseWithStates = async (
+  credentialsAndStates: { signedVc: string; state: string }[],
+  header: JWTHeaderParameters = getDefaultJwtHeader(),
+  forcedCredentialSignatures?: string[]
+): Promise<{ mockEVCSData: CredentialStoreIdentityResponse; credentialSignatures: string[] }> => {
+  const vcs = credentialsAndStates.map((c) => ({ state: c.state, vc: c.signedVc, metadata: undefined }));
+  const credentialSignatures =
+    forcedCredentialSignatures ?? credentialsAndStates.map((c) => c.signedVc.split(".").at(2)!);
+  const storedIdentity = createStoredIdentityRecord(...credentialSignatures);
+  return {
+    mockEVCSData: { si: { vc: await sign(header, storedIdentity), metadata: undefined, unsignedVot: "P3" }, vcs },
+    credentialSignatures,
+  };
+};
+
+export const createCredentialStoreIdentityResponse = async (
+  signedVcs: string[],
+  header: JWTHeaderParameters = getDefaultJwtHeader(),
+  forcedCredentialSignatures?: string[]
+) =>
+  createCredentialStoreIdentityResponseWithStates(
+    signedVcs.map((vc) => ({ signedVc: vc, state: CURRENT })),
+    header,
+    forcedCredentialSignatures
+  );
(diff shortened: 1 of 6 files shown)


src/commons/tests/validate-records.test.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    const header = getDefaultJwtHeader("ES256", "did:web:DISALLOWED.CONTROLLER#f5fe5d2a-9eb6-4819-8c46-723e3a21565a");
+    const { mockEVCSData } = await createCredentialStoreIdentityResponse([], header);
+
+    const result = await validateCryptography(header.kid!, mockEVCSData);
+    expect(result).toEqual({ kidValid: false, signatureValid: false });
+  });
+
+  it("signatureValid false when the SI VC signature does not verify", async () => {
+    const misSignedStoredIdentity =
```
chrisclayson, line 68: Possibly just me, but I wondered why `mis` for a while (the I realised mis-signed 🤦🏻 ) ... might be worth calling the `incorrectlySignedIdentity` or `identityWithBadSignature`?

src/commons/tests/validate-records.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      "eyJhbGciOiJFUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImRpZDp3ZWI6YXBpLmlkZW50aXR5LmRldi5hY2NvdW50Lmdvdi51ayNmNWZlNWQyYS05ZWI2LTQ4MTktOGM0Ni03MjNlM2EyMTU2NWEifQ.eyJzdWIiOiJ1c2VyLXN1YiIsInZvdCI6IlAyIiwidnRtIjpbXX0.-jy9iwsn6uDzr6b3mk0JJZ4NdUf8z3O3ldBbbXKAAtxMH3TIMlBm5u2bI4I1qHrWk1BL2k8muKLV-VIUeych1A";
```
chrisclayson, line 69: Rather than hard code a value here, could you not introduce a helper that generates a dummy private key (in memory) and uses that to produce a signed JWT?

Happy to leave as-is for now, but I would like to revisit this sort of thing in line with the test-credentials library that @chris-dco started writing.

chrisclayson (approved):
Left a couple of non-blocking comments

src/commons/tests/validate-records.test.ts
joe-mcgarry: Good call! I've added a helper function (`generateDummyKey()`) and made the key much easier to read and recognise as a dummy - unless we'd prefer a more realistic looking one?

src/commons/tests/validate-records.test.ts
joe-mcgarry: Yes, good call! Updated that now

shared-test/credential-store-utilities.ts
```diff
@@ -0,0 +1,50 @@
...
+  header: JWTHeaderParameters = getDefaultJwtHeader(),
+  forcedCredentialSignatures?: string[]
+) =>
+  createCredentialStoreIdentityResponseWithStates(
+    signedVcs.map((vc) => ({ signedVc: vc, state: CURRENT })),
+    header,
+    forcedCredentialSignatures
+  );
+
+export const generateDummyKey = (): string => {
```
chrisclayson: I think we should rename this, as its not a key we're generating, its a JWT with a bad signature.

shared-test/credential-store-utilities.ts
```diff
@@ -0,0 +1,50 @@
...
+  return "test.bad.sig";
```
chrisclayson: I think this will work, I was thinking we could make the header and body be realistic (we've got methods already in this module to generate those, and append a bad signature. 

At worst you could do something like:
```
return createSignedIdentityCheckCredentialJWT(issuer) + 'aaa';
```

chrisclayson: Though that does incur the cost of signing it before wrecking the signature 🤔

src/handlers/post-phase2-user-identity-handler/tests/post-phase2-user-identity-handler.test.ts
```diff
@@ -628,69 +507,69 @@ describe("user-identity-handler max_vot", () => {
...
-    nbf: nbfSeconds,
-    sub: "sdf",
-    vc: {
-      evidence: [{}],
-      type: ["VerifiableCredential", "IdentityCheckCredential"],
-    },
-  };
-  return await sign(getDefaultJwtHeader(), credential);
-};
+// type StoredIdentityResponse = {
```
chrisclayson: Did you mean to leave this lot commented out?

joe-mcgarry: I did not mean to! Thought I'd deleted it, but apparently not 😬

shared-test/credential-store-utilities.ts
joe-mcgarry: I don't know if it's the approach you'd want but going by the above comment about signing, what I've done is create `createInvalidIdentityCheckCredentialJWT` which is ultimately a replica of `createSignedIdentityCheckCredentialJWT` only without the call to `sign()`. I figured it was worth a shot at, until I can maybe pick your brain about a better approach later

chrisclayson (approved):

shared-test/credential-store-utilities.ts
joe-mcgarry: For everyone's benefit, this has been removed and replaced by `createInvalidIdentityCheckCredentialJWT()`

